PR #21, opened 2020-03-20
Add install docs

zenmonkeykstop:
Adds installation docs page in Admin section.


docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+
+- In the the ``vault`` file manager, select **+ Other Locations**., then click the persistent volume's listiing in the right panel. It will be named ```N GB encrypted``, where N is the size of the persistent volume. Enter the SVS persistent volume passphrase to unlock and mount it. 
+
+- Open an administration terminal via **Q > Terminal Emulator**, and run the following command:
+
+  .. code-block:: sh
+
+    qvm-run --pass-io vault "gpg --homedir /run/media/user/TailsData/gnupg --export-secret-keys --armor <SVSFingerprint>" > /tmp/sd-journalist.sec
+
+  where ``<SVSFingerprint>`` is the submission key fingerprint. This will copy the submission key in ASCII format to a temporary file in dom0, ``/tmp/sd-journalist.sec``. Verify the that the file starts with ``-----BEGIN PGP PRIVATE KEY BLOCK-----`` using the command:
```
eloquence: should note formatting for fingerprint, e.g., without spaces

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+
+    qvm-run --pass-io vault "cat /run/media/user/TailsData/Persistent/securedrop/install_files/ansible-base/app-journalist.auth_private" > /tmp/journalist.txt
+
+  If your instance uses legacy v2 onion services, use the following command instead:
+
+  .. code-block:: sh
+
+    qvm-run --pass-io vault "cat /run/media/user/TailsData/Persistent/securedrop/install_files/ansible-base/app-journalist.aths" > /tmp/journalist.txt
+
+- Verify that the `/tmp/journalist` file on ``dom0`` contains valid configuration information using the command ``cat /tmp/journalist.txt`` in the administration terminal.
```
eloquence: `/tmp/journalist` -> `/tmp/journalist.txt`

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+    dnf download securedrop-workstation-dom0-config
+
+- In the ``dom0`` administration terminal, run the following command to transfer the RPM package to dom0 and install SecureDrop Workstation:
+
+  .. code-block:: sh
+
+    qvm-run --pass-io work 'cat /home/user/securedrop-workstation-dom0-config-<versionNumber>-1.fc25.noarch.rpm' > securedrop-workstation.rpm
+    sudo dnf install securedrop-workstation.rpm
+
+  where ``<versionNumber>`` is the current SecureDrop Workstation release version. When prompted, press **Y** to install the package.
```
eloquence: Since they've just downloaded it, we can instruct them to note down the version number after the download.

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+   
+    cd /usr/share/securedrop-workstation-dom0-config
+    sudo cp config.json.example config.json
+
+- The ``config.json`` file must be updated with the correct values for your instance. Open it with root privileges in a text editor and update the following fields' values:
+
+  - **submission_key_fpr**: use the value of the submission key fingerprint
+  - **hidserv.hostname**: use the hostname of the Journalist Interface, including the ``.onion`` TLD
+  - **hidserv.key**: use the value of the v2 HidServAuth token for the Journalist Interface, or the v3 private authorization key value if your SecureDrop instance uses v3 onion services
+  - **environment**: use the value ``prod``
```
eloquence: Since that is the default, and there are other variables in the file, I would omit this from the list, and simply say that the other variables can be kept at their default values.

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+
+You can find the values for the **hidserv.*** fields in the ``/tmp/journalist.txt`` file that you created in ``dom0`` earlier.
```
eloquence: The format of `app-journalist.auth_private` is non-obvious, because it does not include the `.onion` TLD. I think we have to explain that here to avoid confusion.

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+  .. code-block:: sh
+
+    securedrop-admin --apply
+
+This command will take a considerable amount of time, as it sets up multiple VMs and installs supporting packages. When it completes, your SecureDrop Workstation is finally ready to use!
+
+Test the Workstation
+--------------------
+
+To start the SecureDrop Workstation client, double-click the SecureDrop desktop icon that was set up by the previous command. The client launcher will start and check for updates. The system should be up-to-date and no updates should be required, but if updates are available follow the instructions in the launcher to apply them.
```
eloquence: launcher -> preflight updater (launcher is an internal name)

docs/admin/install.rst
zenmonkeykstop: Security implications if this, say, gets changed to dev or staging in the sample file in a later release. I would prefer to be explicit.

docs/admin/install.rst
zenmonkeykstop: yup, that came up during rehearsal. Will provide extra detail.

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+
+Install QubesOS
+---------------
+To begin the Qubes installation, connect the Qubes install USB to your target computer and boot from it. You may need to bring up a a boot menu at startup to do so - on Lenovo laptops, for example, you can do so by pressing **F12** on boot.
+
+Follow the `installation documentation <https://www.qubes-os.org/doc/installation-guide/>`_ to install Qubes on your computer, ensuring that you:
+
+- Use all available storage space for the installation (as the computer should be dedicated to SecureDrop Workstation).
+- Set a strong FDE passphrase - a 6-word Diceware passphrase is recommended.
+- Create an administrative account with a strong password. This documentation assumes the username is ``user``.
```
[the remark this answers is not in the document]
eloquence: I think we should discuss the value of a "strong password" for the administrative password in the context of our threat model. Once you have the FDE passphrase, you can boot into single user mode and exfiltrate secrets. A complex password for the main Qubes account adds significant burden to the admin user who will have to type this password in e.g. from a phone, or memorize it. Does it add commensurate value?

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+
+Prerequisites
+-------------
+In order to install SecureDrop Workstation and configure it to use an existing SecureDrop instance, you will need the following:
+
+- A Qubes-compatible computer with at least 16GB of RAM (32 GB is recommended). SecureDrop Workstation has mainly been tested against Lenovo 6th-gen T470 and X1 models, but there are no formally supported hardware recommendations at this time. See Qubes' `Hardware Compatibility List <https://www.qubes-os.org/hcl/>`_ for more options.
+- Qubes installation medium - this guide assumes the use of a USB 3.0 stick. (Qubes may also be installed via optical media, which may make more sense depending on your `security concerns <https://www.qubes-os.org/doc/install-security/>`_.
+- The SecureDrop instance's Admin Workstation and Secure Viewing Station USBs, and the full GPG fingerprint of the submission key.
+- A working computer (Linux is recommended and assumed in this guide) to use for verification and creation of the Qubes installation medium.
+- A password manager or other system to generate and store strong passphrases for Qubes FDE and user accounts.
```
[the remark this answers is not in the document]
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+
+Once the BIOS is up-to-date, boot into the BIOS setup utility and verify that:
+
+- clock settings are correct
+- virtualization support is enabled
+  - for Intel-based devices, **Intel VT-d** and **Intel VT-x** should be enabled
+  - for AMD-based devices, **AMD-VI** and **AMD-V** should be enabled
+- SecureBoot is disabled.
+
+If the Qubes hardware compatibility list entry for your computer recommends the use of Legacy Mode for boot, change that setting in the BIOS as well, otherwise choose UEFI or leave it on the default.
```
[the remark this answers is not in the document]
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+Follow the linked instructions to `verify the ISO <https://www.qubes-os.org/security/verifying-signatures/#how-to-verify-qubes-iso-signatures>`_.
+
+Once you've verified the ISO, copy it to your installation medium - for example, if using Linux and a USB stick, using the command:
+
+.. code-block:: sh
+ 
+  sudo dd if=Qubes-R4.0.3-x86_64.iso of=/dev/sdX bs=1048576 && sync
+
+where ``if`` is set to the path to your downloaded ISO file and ``of`` is set to
+the block device corresponding to your USB stick. Make sure to verify that you have the correct device name using, for example, the ``lsblk`` command.
```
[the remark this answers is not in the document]
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+-------------
+In order to install SecureDrop Workstation and configure it to use an existing SecureDrop instance, you will need the following:
+
+- A Qubes-compatible computer with at least 16GB of RAM (32 GB is recommended). SecureDrop Workstation has mainly been tested against Lenovo 6th-gen T470 and X1 models, but there are no formally supported hardware recommendations at this time. See Qubes' `Hardware Compatibility List <https://www.qubes-os.org/hcl/>`_ for more options.
+- Qubes installation medium - this guide assumes the use of a USB 3.0 stick. (Qubes may also be installed via optical media, which may make more sense depending on your `security concerns <https://www.qubes-os.org/doc/install-security/>`_.
+- The SecureDrop instance's Admin Workstation and Secure Viewing Station USBs, and the full GPG fingerprint of the submission key.
+- A working computer (Linux is recommended and assumed in this guide) to use for verification and creation of the Qubes installation medium.
+- A password manager or other system to generate and store strong passphrases for Qubes FDE and user accounts.
+
+A basic knowledge of the Qubes OS is helpful.
```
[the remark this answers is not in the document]
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+ 
+  sudo dd if=Qubes-R4.0.3-x86_64.iso of=/dev/sdX bs=1048576 && sync
+
+where ``if`` is set to the path to your downloaded ISO file and ``of`` is set to
+the block device corresponding to your USB stick. Make sure to verify that you have the correct device name using, for example, the ``lsblk`` command.
+
+
+Install QubesOS
+---------------
+To begin the Qubes installation, connect the Qubes install USB to your target computer and boot from it. You may need to bring up a a boot menu at startup to do so - on Lenovo laptops, for example, you can do so by pressing **F12** on boot.
```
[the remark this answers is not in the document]
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+
+  This will update ``dom0``, the administration VM, with some required fixes to properly update other VMs. When prompted, type **Y** to apply the updates.
+
+
+- Once the ``dom0`` update is complete, select **Q > System Tools > Qubes Update** to update other VMs. in the ``[Dom0] Qubes Updater`` window, first check ``Enable updates for qubes without known available updates``, then check all entries in the list above. Then, click **Next**. The system's VMs will be updated sequentially - this may take some time. When the updates are complete, click **Finish**.
+
+
+Copy the submission key
+-----------------------
+In order to decrypt submissions, your SecureDrop Workstation will need a copy of the secret key from your SecureDrop instance's SVS. To protect this key and preserve the air gap, you will need to connect the SVS USB to a Qubes VM with no network access, and copy it from there to ``dom0``:
```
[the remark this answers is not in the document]
zenmonkeykstop: Used in prerequisite, so added acronym there.

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+
+- First, use the network manager widget in the upper right panel to disable your network connection. These instructions refer to the ``vault`` VM, which has no network access by default, but if the SVS USB is attached to another VM by mistake, this will offer some protection against exfiltration.
+
+- Next, choose **Q > Domain: vault > vault: Files** to open the file manager in the ``vault`` VM.
+
+- Connect the SVS USB to a USB port on the Qubes computer, then use the devices widget in the upper right panel to attach it to the ``vault`` VM. There will be 3 listings for the USB in the widget: one for the base USB, one for the Tails partition on the USB, labeled ``Tails``, and a 3rd unlabeled listing, for the persistent volume. Choose the third listing.
+
+- In the the ``vault`` file manager, select **+ Other Locations**., then click the persistent volume's listiing in the right panel. It will be named ```N GB encrypted``, where N is the size of the persistent volume. Enter the SVS persistent volume passphrase to unlock and mount it.
```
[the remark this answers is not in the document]
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
zenmonkeykstop: done.

docs/admin/install.rst
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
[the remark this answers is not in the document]
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+To begin the Qubes installation, connect the Qubes install USB to your target computer and boot from it. You may need to bring up a a boot menu at startup to do so - on Lenovo laptops, for example, you can do so by pressing **F12** on boot.
+
+Follow the `installation documentation <https://www.qubes-os.org/doc/installation-guide/>`_ to install Qubes on your computer, ensuring that you:
+
+- Use all available storage space for the installation (as the computer should be dedicated to SecureDrop Workstation).
+- Set a strong FDE passphrase - a 6-word Diceware passphrase is recommended.
+- Create an administrative account with a strong password. This documentation assumes the username is ``user``.
+
+  .. note:: Qubes is not intended to have multiple user accounts, so this account name and password will be shared by all SwcureDrop Workstation users.
+
```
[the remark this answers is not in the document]
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+Once the installation is complete, you will be prompted to reboot into Qubes. Reboot, removing the install USB when the computer restarts. 
+
+You will be prompted to enter the FDE passphrase set during installation. After the disk is unlocked and Qubes starts, you will be prompted to complete the initial setup. Click the Qubes OS icon, then accept the default options and click **Done**. Finally, click **Finish Configuration** to set up the default system TemplateVMs and AppVMs.
```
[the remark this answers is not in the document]
zenmonkeykstop: added in a note.

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+
+Copy the submission key
+-----------------------
+In order to decrypt submissions, your SecureDrop Workstation will need a copy of the secret key from your SecureDrop instance's SVS. To protect this key and preserve the air gap, you will need to connect the SVS USB to a Qubes VM with no network access, and copy it from there to ``dom0``:
+
+- First, use the network manager widget in the upper right panel to disable your network connection. These instructions refer to the ``vault`` VM, which has no network access by default, but if the SVS USB is attached to another VM by mistake, this will offer some protection against exfiltration.
+
+- Next, choose **Q > Domain: vault > vault: Files** to open the file manager in the ``vault`` VM.
+
+- Connect the SVS USB to a USB port on the Qubes computer, then use the devices widget in the upper right panel to attach it to the ``vault`` VM. There will be 3 listings for the USB in the widget: one for the base USB, one for the Tails partition on the USB, labeled ``Tails``, and a 3rd unlabeled listing, for the persistent volume. Choose the third listing.
```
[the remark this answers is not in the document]
zenmonkeykstop: This would be better handled in a generic qubes intro, or via screenshots imo.

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+- In the the ``vault`` file manager, select **+ Other Locations**., then click the persistent volume's listiing in the right panel. It will be named ```N GB encrypted``, where N is the size of the persistent volume. Enter the SVS persistent volume passphrase to unlock and mount it. 
+
+- Open an administration terminal via **Q > Terminal Emulator**, and run the following command:
+
+  .. code-block:: sh
+
+    qvm-run --pass-io vault "gpg --homedir /run/media/user/TailsData/gnupg --export-secret-keys --armor <SVSFingerprint>" > /tmp/sd-journalist.sec
+
+  where ``<SVSFingerprint>`` is the submission key fingerprint. This will copy the submission key in ASCII format to a temporary file in dom0, ``/tmp/sd-journalist.sec``. Verify the that the file starts with ``-----BEGIN PGP PRIVATE KEY BLOCK-----`` using the command:
+
```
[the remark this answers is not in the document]
zenmonkeykstop: Updated command to only display first line instead.

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+  .. code-block:: sh
+
+    qvm-run --pass-io vault "cat /run/media/user/TailsData/Persistent/securedrop/install_files/ansible-base/app-journalist.auth_private" > /tmp/journalist.txt
+
+  If your instance uses legacy v2 onion services, use the following command instead:
+
+  .. code-block:: sh
+
+    qvm-run --pass-io vault "cat /run/media/user/TailsData/Persistent/securedrop/install_files/ansible-base/app-journalist.aths" > /tmp/journalist.txt
+
```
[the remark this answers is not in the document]
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+  .. code-block:: sh
+
+    dnf download securedrop-workstation-dom0-config
+
+- In the ``dom0`` administration terminal, run the following command to transfer the RPM package to dom0 and install SecureDrop Workstation:
+
+  .. code-block:: sh
+
+    qvm-run --pass-io work 'cat /home/user/securedrop-workstation-dom0-config-<versionNumber>-1.fc25.noarch.rpm' > securedrop-workstation.rpm
```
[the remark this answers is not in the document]
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+
+  .. code-block:: sh
+   
+    cd /usr/share/securedrop-workstation-dom0-config
+    sudo cp config.json.example config.json
+
+- The ``config.json`` file must be updated with the correct values for your instance. Open it with root privileges in a text editor and update the following fields' values:
+
+  - **submission_key_fpr**: use the value of the submission key fingerprint
+  - **hidserv.hostname**: use the hostname of the Journalist Interface, including the ``.onion`` TLD
```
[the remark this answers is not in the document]
zenmonkeykstop: added a note to address this.

docs/admin/install.rst
```diff
@@ -0,0 +1,243 @@
...
+  - **hidserv.key**: use the value of the v2 HidServAuth token for the Journalist Interface, or the v3 private authorization key value if your SecureDrop instance uses v3 onion services
+  - **environment**: use the value ``prod``
+
+You can find the values for the **hidserv.*** fields in the ``/tmp/journalist.txt`` file that you created in ``dom0`` earlier.
+
```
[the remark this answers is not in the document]
zenmonkeykstop: I dunno, this is basic unixy/Xy stuff and there's nothing Qubesy in play here. Inclined to leave it. If we have to pitch these docs at folks without CLI experience, there's a lot more to add, and managing the workstation may not be the best place for them to learn this.

docs/admin/install.rst
```diff
@@ -0,0 +1,265 @@
...
+Once you've verified the ISO, copy it to your installation medium - for example, if using Linux and a USB stick, using the command:
+
+.. code-block:: sh
+ 
+  sudo dd if=Qubes-R4.0.3-x86_64.iso of=/dev/sdX bs=1048576 && sync
+
+where ``if`` is set to the path to your downloaded ISO file and ``of`` is set to
+the block device corresponding to your USB stick. 
+
+.. caution:: Make sure to verify that you have the correct device name using, for example, the ``lsblk`` command.
```
eloquence: I would be even more explicit here. "All data on the target device will be overwritten." or something to that effect.

docs/admin/install.rst
```diff
@@ -0,0 +1,265 @@
...
+Follow the linked instructions to `verify the ISO <https://www.qubes-os.org/security/verifying-signatures/#how-to-verify-qubes-iso-signatures>`_.
+
+Once you've verified the ISO, copy it to your installation medium - for example, if using Linux and a USB stick, using the command:
+
+.. code-block:: sh
+ 
+  sudo dd if=Qubes-R4.0.3-x86_64.iso of=/dev/sdX bs=1048576 && sync
+
+where ``if`` is set to the path to your downloaded ISO file and ``of`` is set to
+the block device corresponding to your USB stick.
```
eloquence: If we duplicate these instructions, I would include the same caveat here that Qubes does:
> Make sure to write to the entire device (e.g., /dev/sdc) rather than just a single partition (e.g., /dev/sdc1).

docs/admin/install.rst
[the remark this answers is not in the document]
eloquence: I think earlier, when we instruct folks to create a file, we can mention that both `vi` and `nano` are installed in `dom0`, just so folks know what options are available to them.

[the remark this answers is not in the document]
zenmonkeykstop: File being created in `work`, which doesn't have `nano`! (as @ro found out.) I'm not against mentioning editors here, but again, it's a question of intended audience.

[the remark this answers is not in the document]
eloquence: I think we can assume comfort on the CLI but I don't think we should assume comfort with `vi`, so I do think it's worth including a tip about what works where, and maybe even a brief one-liner for how to install `nano` in `work` temporarily.

docs/admin/install.rst
```diff
@@ -0,0 +1,265 @@
+Installing and Configuring SecureDrop Workstation
+=================================================
+
+.. include:: ../includes/top-warning.rst
+
+Prerequisites
+-------------
+In order to install SecureDrop Workstation and configure it to use an existing SecureDrop instance, you will need the following:
+
+- A Qubes-compatible computer with at least 16GB of RAM (32 GB is recommended). SecureDrop Workstation has mainly been tested against Lenovo 6th-gen T470 and X1 models, but there are no formally supported hardware recommendations at this time. See Qubes' `Hardware Compatibility List <https://www.qubes-os.org/hcl/>`_ for more options.
```
eloquence: We've in fact recommended the T480 for all pilot participants; if we want all folks who install SecureDrop Workstation to update the BIOS, then I think we need to be a bit more explicit about how to do that; the support site is not helpful. Should folks keep a pre-existing Windows partition in place to perform an update? Should they use the bootable CD or the Linux binary? The more we can take the guesswork out of it for the primary hardware platform we're installing on, the better, IMO.

eloquence: Steps that worked for me on T480 for BIOS update:
1. Download .iso image from https://pcsupport.lenovo.com/us/en/products/laptops-and-netbooks/thinkpad-t-series-laptops/thinkpad-t480-type-20l5-20l6/downloads/ds502355
2. Use `geteltorito` to create bootable image
3. Use `dd` to create bootable USB drive
4. Boot from USB and follow instructions.

docs/admin/install.rst
```diff
@@ -0,0 +1,265 @@
...
+Once the update check is complete, the client will launch. Log in using an existing journalist account and verify that sources are listed and submissions can be downloaded, decrypted, and viewed.
+
+Clean up
+--------
+
+Once you've confirmed that the client is working correctly, delete the files containing the key and Journalist Interface details from ``dom0``, using the following command:
+
+.. code-block:: sh
+  
+  rm /tmp/{sd-journalist.sec journalist.txt}
```
emkll: This will be cleared on reboot. we can also use `shred` here, However since it's an SSD it's unclear when this data will really be deleted.

docs/admin/install.rst
```diff
@@ -0,0 +1,265 @@
...
+
+Uninstall SecureDrop Workstation
+--------------------------------
+To uninstall SecureDrop Workstation, open a ``dom0`` terminal and run the following command:
+
+.. code-block:: sh
+  
+  securedrop-admin --uninstall
+
+This will remove all associated VMs and configuration details, and uninstall the ``dom0`` SecureDrop Workstation package.
```
emkll: is https://github.com/freedomofpress/securedrop-workstation/issues/505 worth noting?

docs/admin/install.rst
```diff
@@ -0,0 +1,265 @@
...
+  rm /tmp/{sd-journalist.sec journalist.txt}
+
+
+Uninstall SecureDrop Workstation
+--------------------------------
+To uninstall SecureDrop Workstation, open a ``dom0`` terminal and run the following command:
+
+.. code-block:: sh
+  
+  securedrop-admin --uninstall
```
emkll: Instance secrets will remain in /usr/share/securedrop-workstation-dom0-config until manually deleted, per https://github.com/freedomofpress/securedrop-workstation/blob/cfceca0a7152efdf9f06bf8f0bfe0c9f8d95d689/scripts/securedrop-admin.py#L82

docs/admin/known_issues.rst
```diff
@@ -33,3 +33,4 @@ Current known issues
   source is disabled for more than a minute, they must be flagged for reply in
   the *Journalist Interface* - see the `SecureDrop "Flag for Reply"
   documentation <https://docs.securedrop.org/en/stable/journalist.html#flag-for-reply>`_ for more information.
+- Currently, only app-based two-factor authentication is supported.
```
emkll: TOTP?

docs/admin/install.rst
eloquence: IMO if we expect that warnings will be shown then we should note it here.

eloquence:
Regarding the recommendation for the default user password:
- Any user with FDE passphrase can boot into single user mode, so this password provides limited protection against system compromise.
- It does provide some weak protection in the case where the laptop is left unattended, or against naive compromise attempts.
- User will have to type this password every time they log in and will never be able to copy/paste it, so a password like `@LK$#\x<L:#$` seems highly undesirable.

With these tradeoffs in mind, I think we should use language like this:
> We recommend setting a strong password or passphrase that you are comfortable typing repeatedly.

Qubes itself provides a strength indicate while you set the password, so I think this is sufficient guidance. Thoughts?

docs/admin/install.rst
```diff
@@ -0,0 +1,265 @@
...
+
+Install Qubes OS
+----------------
+To begin the Qubes installation, connect the Qubes install USB to your target computer and boot from it. You may need to bring up a boot menu at startup to do so - on Lenovo laptops, for example, you can do so by pressing **F12** on boot.
+
+Follow the `installation documentation <https://www.qubes-os.org/doc/installation-guide/>`_ to install Qubes on your computer, ensuring that you:
+
+- Use all available storage space for the installation (as the computer should be dedicated to SecureDrop Workstation).
+- Set a strong FDE passphrase - a 6-word Diceware passphrase is recommended.
+- Create an administrative account named `user` with a strong password.
```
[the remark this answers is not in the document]
zenmonkeykstop: this was to introduce some consistency and reduce the possibility for confusion elsewhere - if you're doing stuff on the system-created VMs like `work`, they have `user` as the default account.

docs/admin/install.rst
eloquence: `/tmp` is a tmpfs so uses volatile memory (or swap in extreme circumstances)

docs/admin/install.rst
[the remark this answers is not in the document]
eloquence: I think a screenshot is needed here, especially because the widget lists USB devices both as block devices and as physical devices, which can be very confusing.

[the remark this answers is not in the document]
eloquence: I've taken two screenshots; will upload shortly.

docs/admin/install.rst
```diff
@@ -0,0 +1,265 @@
...
+
+- Connect the SVS USB to a USB port on the Qubes computer, then use the devices widget in the upper right panel to attach it to the ``vault`` VM. There will be 3 listings for the USB in the widget: one for the base USB, one for the Tails partition on the USB, labeled ``Tails``, and a 3rd unlabeled listing, for the persistent volume. Choose the third listing.
+
+- In the the ``vault`` file manager, select **+ Other Locations**., then click the persistent volume's listing in the right panel. It will be named ```N GB encrypted``, where N is the size of the persistent volume. Enter the SVS persistent volume passphrase to unlock and mount it. 
+
+- Open an administration terminal via **Q > Terminal Emulator**, and run the following command:
+
+  .. code-block:: sh
+
+    qvm-run --pass-io vault "gpg --homedir /run/media/user/TailsData/gnupg --export-secret-keys --armor <SVSFingerprint>" > /tmp/sd-journalist.sec
```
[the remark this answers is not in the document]
eloquence: No user should have to type a fingerprint by hand. Perhaps we can instruct them to run `--list-keys` first to identify the full key?

docs/admin/install.rst
```diff
@@ -0,0 +1,265 @@
...
+
+- In the ``vault`` file manager, select **+ Other Locations** and eject the TailsData volume, then disconnect the Admin Workstation USB. Shut down the ``vault`` VM using the Qube widget in the upper right panel.
+
+
+Download and install SecureDrop Workstation
+-------------------------------------------
+
+With the key and configuration available in ``dom0``, you're ready to set up SecureDrop Workstation:
+
+- First, re-enable the network connection using the network manager widget.
```
eloquence: For once, copy and paste will actually work; how about we instruct them to visit this documentation in a browser at this point, so they can copy/paste the commands, don't have to type full fingerprint by hand, etc.?

docs/admin/install.rst
```diff
@@ -0,0 +1,265 @@
...
+- Next, start a terminal in the network-attached ``work`` VM, via **Q > Domain:work > work: Terminal**.
+
+- In the ``work`` terminal, run the following commands to download and add the SecureDrop signing key, which is needed to verify the SecureDrop Workstation package:
+
+  .. code-block:: sh
+
+    gpg --keyserver hkps://keys.openpgp.org --recv-key "2224 5C81 E3BA EB41 38B3 6061 310F 5612 00F4 AD77"
+    gpg --armor --export 22245C81E3BAEB4138B36061310F561200F4AD77 | sudo tee /etc/pki/rpm-gpg/RPM-GPG-KEY-securedrop-workstation
+
+- In the ``work`` terminal, create a file ``/etc/yum.repos.d/securedrop-temp.repo`` with the following contents:
```
eloquence: We should specify that `sudo` is required (implicit if we give example with `vi`, or `nano` after installing)

docs/admin/install.rst
```diff
@@ -0,0 +1,265 @@
...
+    baseurl=https://yum.securedrop.org/workstation/dom0/f25
+    name=SecureDrop Workstation Qubes initial install bootstrap
+
+- Download the SecureDrop Workstation config package with the command:
+
+  .. code-block:: sh
+
+    dnf download securedrop-workstation-dom0-config
+
+Note the release version number in the filename, you'll need it below.
```
eloquence: I would be a bit more explicit about what the command does, e.g., "This will download the RPM package to the current working directory. Note .."

docs/admin/install.rst
```diff
@@ -0,0 +1,265 @@
...
+- Next, create the SecureDrop Workstation configuration file:
+
+  .. code-block:: sh
+   
+    cd /usr/share/securedrop-workstation-dom0-config
+    sudo cp config.json.example config.json
+
+- The ``config.json`` file must be updated with the correct values for your instance. Open it with root privileges in a text editor and update the following fields' values:
+
+  - **submission_key_fpr**: use the value of the submission key fingerprint
```
eloquence: They can get the fingerprint in friendly copy/paste format by running `gpg --with-fingerprint --with-colons sd-journalist.sec`, IMO worth pointing out here so they don't have to go digging for it again, or re-type it.

docs/admin/install.rst
```diff
@@ -0,0 +1,265 @@
...
+  .. code-block:: none
+  
+   ONIONADDRESS:descriptor:x25519:AUTHTOKEN
+
+
+- Finally, in the ``dom0`` terminal, run the command:
+
+  .. code-block:: sh
+
+    securedrop-admin --apply
```
eloquence: We could have them run `--validate` first to ensure correct format, or we can just point out that this command will report any syntax errors detected in the configuration.

eloquence:
Screenshots for the attach flow:
ws2px2e
gOfDuFA

docs/admin/install.rst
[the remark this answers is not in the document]
zenmonkeykstop: TBH I think it adds more opportunity for error to say to users "this needs a password that's strong, this not so much", as they now have to decide what that means. Given password managers exist and we encourage their use, I would suggest that we keep things binary; either we recommend a strong passphrase, or in situations where, for example, stuff is behind an FDE passphrase, we recommend no passphrase.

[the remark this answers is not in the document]
zenmonkeykstop: In this case, as we don't know if folks will treat the workstation like the SVS physical-security-wise, IMO it's probably worth setting a passphrase to guard against shoulder surfers or unattended machines. (This does not guard against the workstation storage being read if left unlocked.)

docs/admin/install.rst
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
```diff
@@ -0,0 +1,265 @@
...
+-------------------------------------
+Before beginning the Qubes installation, make sure that your Qubes-compatible computer's BIOS is updated to the latest available version. Instructions on how to do so will vary by manufacturer or model and can usually be found on their respective support sites, for example `https://pcsupport.lenovo.com/ <https://pcsupport.lenovo.com/>`_ in the case of Lenovo.
+
+Once the BIOS is up-to-date, boot into the BIOS setup utility and verify that:
+
+- clock settings are correct
+- virtualization support is enabled
+  - for Intel-based devices, **Intel VT-d** and **Intel VT-x** should be enabled
+  - for AMD-based devices, **AMD-VI** and **AMD-V** should be enabled
+- SecureBoot is disabled.
```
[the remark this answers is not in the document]
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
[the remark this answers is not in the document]
eloquence: Yeah a passphrase that's easy to type would work, if there's no character limit. Whatever language we use should be explicit enough to avoid them picking a password that's impossible or overly laborious to type since they'll have to unlock screensaver with it repeatedly.

docs/admin/install.rst
```diff
@@ -0,0 +1,265 @@
...
+- Next, configure Tor by selecting the Qubes menu (the **Q** icon in the upper left corner) and selecting **Service: sys-whonix > sys-whonix: Anon Connection Wizard**. In most cases, choosing the default**Connect** option is best. Click **Next**, then **Next** again. Then, if Tor connects successfully, click **Finish**. If Tor fails to connect, make sure your network conection is up and does not filter Tor connections, then try again.
+
+  .. note:: If Tor connections are blocked on your network, you may need to configure Tor to use bridges in order to get a connection. For more information, see the `Anon Connection Wizard <https://www.whonix.org/wiki/Anon_Connection_Wizard>`_ documentation.
+
+
+- Open an administration terminal by selecting **Q > Terminal Emulator**, and enter the following command:
+
+  .. code-block:: sh
+
+   sudo qubes-dom0-update --enablerepo=qubes-dom0-current-testing
```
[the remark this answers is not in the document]
zenmonkeykstop: To work around #485.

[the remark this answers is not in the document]
eloquence: I'd suggest adding a note that this is a temporary setting and should not be used for a real prod install.

docs/admin/install.rst
```diff
@@ -0,0 +1,265 @@
...
+----------------
+To begin the Qubes installation, connect the Qubes install USB to your target computer and boot from it. You may need to bring up a boot menu at startup to do so - on Lenovo laptops, for example, you can do so by pressing **F12** on boot.
+
+Follow the `installation documentation <https://www.qubes-os.org/doc/installation-guide/>`_ to install Qubes on your computer, ensuring that you:
+
+- Use all available storage space for the installation (as the computer should be dedicated to SecureDrop Workstation).
+- Set a strong FDE passphrase - a 6-word Diceware passphrase is recommended.
+- Create an administrative account named `user` with a strong password. 
+
+  .. note:: Qubes is not intended to have multiple user accounts, so this account name and password will be shared by all SecureDrop Workstation users.
```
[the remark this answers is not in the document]
zenmonkeykstop: It's a two-line paragraph that does not provide information pertinent to the installation. In the interests of keeping things simple, I vote against :)

docs/admin/install.rst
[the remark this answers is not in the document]
zenmonkeykstop: Always looking out for the user :light: :bike:.
:heavy_check_mark:

docs/admin/install.rst
```diff
@@ -0,0 +1,265 @@
...
+  .. code-block:: sh
+
+    head -n 1 /tmp/sd-journalist.sec
+
+- in the ``vault`` file manager, select **+ Other Locations** and eject the TailsData volume, then disconnect the SVS USB. 
+
+Copy Journalist Interface details 
+---------------------------------
+
+SecureDrop Workstation connects to your SecureDrop instance's API via the Journalist Interface. In order to do so, it will need the Journalist Interface address and authorization info. This can be copied from the Admin Workstation via the following steps:
```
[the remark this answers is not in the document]
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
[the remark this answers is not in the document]
zenmonkeykstop: also added note re: copy-paste.

docs/admin/install.rst
```diff
@@ -0,0 +1,265 @@
...
+
+- Connect the Admin Workstation USB to a USB port on the Qubes computer, then use the devices widget in the upper right panel to attach it to the ``vault`` VM. There will be 3 listings for the USB in the widget: one for the base USB, one for the Tails partition on the USB, labeled ``Tails``, and a 3rd unlabeled listing, for the persistent volume. Choose the third listing.
+
+- In the the ``vault`` file manager, select **+ Other Locations**., then click the persistent volume's listing in the right panel. It will be named ```N GB encrypted``, where N is the size of the persistent volume. Enter the Admin Workstation persistent volume passphrase to unlock and mount it. 
+
+- Copy the Journalist Interface configuration file to `dom0`. If your SecureDrop instance uses v3 onion services, use the following command:
+
+  .. code-block:: sh
```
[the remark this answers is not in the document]
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
```diff
@@ -0,0 +1,265 @@
...
+- In the ``dom0`` administration terminal, run the following command to transfer the RPM package to dom0 and install SecureDrop Workstation:
+
+  .. code-block:: sh
+
+  qvm-run --pass-io work "cat /home/user/securedrop-workstation-dom0-config-<versionNumber>-1.fc25.noarch.rpm" > securedrop-workstation.rpm
+    sudo dnf install securedrop-workstation.rpm
+
+  where ``<versionNumber>`` is the current SecureDrop Workstation release version. When prompted, press **Y** to install the package.
+ 
+Configure SecureDrop Workstation
```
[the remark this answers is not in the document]
zenmonkeykstop: We're not explicitly asking them to reboot, so it seems superfluous.

docs/admin/install.rst
```diff
@@ -0,0 +1,265 @@
...
+   ONIONADDRESS:descriptor:x25519:AUTHTOKEN
+
+
+- Finally, in the ``dom0`` terminal, run the command:
+
+  .. code-block:: sh
+
+    securedrop-admin --apply
+
+This command will take a considerable amount of time, as it sets up multiple VMs and installs supporting packages. When it completes, your SecureDrop Workstation is finally ready to use!
```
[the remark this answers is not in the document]
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
zenmonkeykstop: Imma delete that section in that case.

docs/admin/install.rst
zenmonkeykstop: This is a bug that we should fix. We should not expect users to interpret error messages to figure out if they're "okay" ones.

docs/admin/install.rst
zenmonkeykstop: Added a para addressing this.

docs/admin/install.rst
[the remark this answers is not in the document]
zenmonkeykstop: My bad, this file is being updated in dom0, which does have nano - I was thinknig of the repo config file in work. Updating accordingly, with a mental note to revisit this when securedrop-admin groks aths/.auth_private files.

docs/admin/install.rst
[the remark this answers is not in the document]
zenmonkeykstop: Diceware passwords are supposed to be less hassle to type.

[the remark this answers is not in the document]
zenmonkeykstop: OK updated note accordingly

docs/admin/install.rst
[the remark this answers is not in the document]
zenmonkeykstop: agreed.

docs/admin/install.rst
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
zenmonkeykstop: nice. 
:heavy_check_mark:

docs/admin/install.rst
zenmonkeykstop: :heavy_check_mark:

docs/admin/known_issues.rst
zenmonkeykstop: updated.

emkll:
Provided the finding/changes in https://github.com/freedomofpress/securedrop-workstation/pull/510 is valid, we should make an update to these docs as well.

docs/admin/install.rst
eloquence: IMO still worth addressing explicitly, there's a lot of manual tippety-typing in this section otherwise.

docs/admin/install.rst
zenmonkeykstop: BIOS updates have been split out to a separate hardware page in latest commit.

docs/admin/hardware.rst
```diff
@@ -0,0 +1,65 @@
...
+The ThinkPad T480 is a viable option for SecureDrop Workstation, and also has the advantage of being used heavily by the SecureDrop development team. If you plan to use it, you should follow the instructions below to ensure that the BIOS is up to date before proceeding with the installation:
+
+.. _t480_bios:
+
+Upgrading the T480 BIOS
+~~~~~~~~~~~~~~~~~~~~~~~
+
+The instructions below assume the use of a Linux-based computer for the creation of a BIOS upgrade USB. To upgrade the T480 BIOS:
+
+- Locate the model number of the T480 - it can often be found on a sticker on the underside of the laptop, or via the ``Main`` tab in Thinkpad Setup (accessed by pressing **Enter** on startup).
```
eloquence: "Model number" is ambiguous in the Lenovo context; "machine type" is the string they will need, which is a substring of what is called the "machine type model" in the BIOS screen. I think we should explicitly call it the "machine type" and give examples like `20L5` and `20L6`. Few things are more frustrating than trying to bash strings into the support portal search that produce no results, which is what will happen if you try to search for the full "machine type model" or other strings you may find on a sticker.

An annotated sticker like https://www.cpsc.gov/s3fs-public/styles/large/public/767_3.png?dwqT2uXDRkPddkqfw9lL4OOosQi6ht_r&itok=BTGkEBQg could be helpful, but I'm noting that my brand-new T480 straight from Lenovo has no stickers.

docs/admin/hardware.rst
```diff
@@ -0,0 +1,65 @@
...
+
+  .. code-block:: sh
+   
+    sha256sum /path/to/downloaded.iso
+
+- Create a USB-bootable version of the ISO using the command:
+
+  .. code-block:: sh
+
+    geteltorito <path/to/CDISO> > usb-bios.iso
```
eloquence: Need to install `genisoimage` if they don't have `geteltorito` (I had to install it on Ubuntu 18.04, haven't tested other distros).

docs/admin/hardware.rst
```diff
@@ -0,0 +1,65 @@
...
+
+- Plug in a USB and check its device name with the ``lsblk`` command - use the root device name below, not a partition (eg. ``/dev/sdc`` instead of ``/dev/sdc1``).
+
+- Write the BIOS update ISO to the USB using the following command:
+
+  .. code-block:: sh
+
+    ``sudo dd if=usb-bios.iso of=/dev/sda bs=1M && sync``
```
eloquence: Sorry, but like in the Qubes docs, I think we'll again have to be a bit more explicit with these instructions, and I would caution against using `/dev/sda` in any example, since it's likely to be their main drive.

docs/admin/hardware.rst
```diff
@@ -0,0 +1,65 @@
...
+
+  Once complete, remove the USB.
+
+- Plug the USB into the T480 and boot it, pressing **F12** on startup. Select the USB's listing in the boot menu.
+
+- Follow the on-screen instructions to update the BIOS, including any mandatory reboots.
```
eloquence: I would add something like:

> Note that the on-screen instructions may refer to a CD instead of a USB drive, which is expected behavior.

docs/admin/hardware.rst
```diff
@@ -0,0 +1,65 @@
...
+- 64-bit Intel or AMD processor with virtualization support
+- a minimum of 16GB RAM (32GB recommended for production use)
+- sufficient disk space for the Qubes OS base install and SecureDrop Workstation VMs (a 128GB or greater SSD is recommended)
+
+More information on hardware compatibility can be found on the `Qubes OS System Requirements <https://www.qubes-os.org/doc/system-requirements/>`_ page, and information on specific systems can be found via the `hardware compatibility list <https://www.qubes-os.org/hcl/>`_.
+
+In order to print submissions, a supported non-networked printer is required. Supported models currently include:
+
+- TBD
+- Also TBD
```
eloquence: Brother HL-L2320D or HP LaserJet Pro M404n printer

docs/admin/hardware.rst
```diff
@@ -0,0 +1,65 @@
...
+
+More printer options will be added in future releases.
+
+Lenovo ThinkPad T480
+--------------------
+The ThinkPad T480 is a viable option for SecureDrop Workstation, and also has the advantage of being used heavily by the SecureDrop development team. If you plan to use it, you should follow the instructions below to ensure that the BIOS is up to date before proceeding with the installation:
```
eloquence: I wouldn't soft-pedal this, we can just say something like "We currently recommend using the Thinkpad T480 for the SecureDrop Workstation, and are using it during testing, development, and quality assurance."

eloquence:
Added some notes to BIOS section; thanks a lot for putting this together and for your patience w/ review, these docs are really looking great already.

docs/admin/hardware.rst
zenmonkeykstop: :heavy_check_mark:

docs/admin/hardware.rst
zenmonkeykstop: added a note.

:heavy_check_mark:

docs/admin/hardware.rst
zenmonkeykstop: :heavy_check_mark:

docs/admin/hardware.rst
zenmonkeykstop: :heavy_check_mark:

docs/admin/hardware.rst
zenmonkeykstop: No other references to "we" in install or hardware docs, only "you". Will reword.

docs/admin/hardware.rst
```diff
@@ -0,0 +1,83 @@
...
+
+.. _t480_bios:
+
+Upgrading the T480 BIOS
+~~~~~~~~~~~~~~~~~~~~~~~
+
+The instructions below assume the use of a Linux-based computer for the creation of a BIOS upgrade USB. To upgrade the T480 BIOS:
+
+- Locate the machine type of the T480 - it be found via the ``Main`` tab in Thinkpad Setup (accessed by pressing **Enter** on startup). For recent T480s, it will start with a string like `20L5` or `20L6`.
+- Visit `<https://support.lenovo.com>`_ in the Linux-based computer. Type the model number found above into the search bar, then press **Enter**.
```
eloquence: model number -> machine type

docs/admin/hardware.rst
```diff
@@ -0,0 +1,83 @@
...
+The ThinkPad T480 is a recommended option for SecureDrop Workstation, as it is being used by the core team for development and testing. If you plan to use it, you should follow the instructions below to ensure that the BIOS is up to date before proceeding with the installation:
+
+.. _t480_bios:
+
+Upgrading the T480 BIOS
+~~~~~~~~~~~~~~~~~~~~~~~
+
+The instructions below assume the use of a Linux-based computer for the creation of a BIOS upgrade USB. To upgrade the T480 BIOS:
+
+- Locate the machine type of the T480 - it be found via the ``Main`` tab in Thinkpad Setup (accessed by pressing **Enter** on startup). For recent T480s, it will start with a string like `20L5` or `20L6`.
```
eloquence: To avoid confusion, I would say something like "it will _be_ a string", not "start with a string", since the full machine type number is _not_ what they want (it won't produce matches on the support portal)

eloquence:
Couple small comments re: the BIOS docs, and would still recommend copy/paste instructions for the steps performed in the `work` VM, but otherwise this LGTM from my end :)

docs/admin/hardware.rst
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
zenmonkeykstop: I'm against it, just because it's a bit of a context switch and I want to keep the amount of things happening on the workstation itself to a minimum.

docs/admin/hardware.rst
zenmonkeykstop: :heavy_check_mark:

docs/admin/hardware.rst
zenmonkeykstop: :heavy_check_mark:

zenmonkeykstop:
Thanks, pushed a rebased version to keep the commit count down :)

docs/admin/install.rst
eloquence: Can you clarify what process you are suggesting? Are you recommending they type in all the commands and file contents? That seems laborious and error-prone, especially since it includes a fingerprint (twice). IMO this is not a minor issue -- copy/paste in Qubes is very much nonstandard, and we're extremely careful about when we suggest using a browser or network-enabled VMs, so conservative users may assume it is _not_ safe to use a browser, which is not the case.

As a simple tip I don't think it adds a significant context switch, e.g.: "If you do not want to manually type the following commands, you can safely open a web browser in the `work` VM and copy and paste directly from this documentation.`

zenmonkeykstop: Ok, but if they choose to do so, then you have to explain in which contexts they can use copy/paste. Every choice and every optional window is extra mental overhead, which increases the chance of people getting lost in the install process. Keeping the instructions as linear as possible (at the cost of having them type some commands) reduces that overhead.

eloquence: In general, I don't think we have the luxury to avoid talking about copy and paste. It's an essential operation in any operating system, and we will need to add a section to the overall docs explaining what does and doesn't work, and why. We will update this as we make adjustments to that policy, which I'm quite certain we will need to do given how conservative it currently is.

In this context, we're asking the user to create configuration files from scratch, import fingerprints, etc. So context-switch or not, I do think we need to give them a recommended option for avoiding having to manually type all those instructions.

zenmonkeykstop: Fair enough, I'll add in a note here.

eloquence:
Looks great, doing a quick final visual review before merging. :)

docs/admin/install.rst
```diff
@@ -0,0 +1,336 @@
...
+Once the initial setup is complete, the login dialog will be displayed. Log in using the username and password set during installation. 
+
+Apply updates to system templates (estimated wait time: 45-60 minutes)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Before installing SecureDrop Workstation, you must set up network and Tor access, then update the system VMs:
+
+- After logging in, use the network manager widget in the upper-right panel to configure your network connection.
+
+- Next, configure Tor by selecting the Qubes menu (the **Q** icon in the upper left corner) and selecting **Service: sys-whonix > sys-whonix: Anon Connection Wizard**. In most cases, choosing the default**Connect** option is best. Click **Next**, then **Next** again. Then, if Tor connects successfully, click **Finish**. If Tor fails to connect, make sure your network conection is up and does not filter Tor connections, then try again.
```
eloquence: `default**Connect**` -> `default **Connect**`

docs/admin/install.rst
```diff
@@ -0,0 +1,336 @@
...
+
+- In the the ``vault`` file manager, select **+ Other Locations**, then click the persistent volume's listing in the right panel. It will be named ``N GB encrypted``, where N is the size of the persistent volume. Enter the SVS persistent volume passphrase to unlock and mount it. 
+
+  |Unlock TailsData|
+
+- Open a ``dom0`` terminal via **Q > Terminal Emulator**, and run the following command to list the SVS submission key details, including its fingerprint:
+
+  .. code-block:: sh
+
+    qvm-run --pass-io vault "gpg --homedir /run/media/user/TailsData/gnupg -K --fingerprint"
```
eloquence: My output includes the encryption subkey fingerprint; I don't think we need to change anything here now, but I think it would be useful as an iteration on these docs to show example output here to ensure users copy the correct fingerprint.

docs/admin/install.rst
```diff
@@ -0,0 +1,336 @@
...
+
+
+Copy Journalist Interface details
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+SecureDrop Workstation connects to your SecureDrop instance's API via the Journalist Interface. In order to do so, it will need the Journalist Interface address and authentication info. As the clipboard from another VM cannot be copied into ``dom0`` directly, follow these steps to copy the file into place:
+
+- Connect the Admin Workstation USB to a USB port on the Qubes computer, then use the devices widget in the upper right panel to attach it to the ``vault`` VM. There will be 3 listings for the USB in the widget: one for the base USB, one for the Tails partition on the USB, labeled ``Tails``, and a 3rd unlabeled listing, for the persistent volume. Choose the third listing.
+
+- In the the ``vault`` file manager, select **+ Other Locations**., then click the persistent volume's listing in the right panel. It will be named ```N GB encrypted``, where N is the size of the persistent volume. Enter the Admin Workstation persistent volume passphrase to unlock and mount it.
```
eloquence: stray `.` after "Other Locations"

docs/admin/install.rst
```diff
@@ -0,0 +1,336 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+SecureDrop Workstation connects to your SecureDrop instance's API via the Journalist Interface. In order to do so, it will need the Journalist Interface address and authentication info. As the clipboard from another VM cannot be copied into ``dom0`` directly, follow these steps to copy the file into place:
+
+- Connect the Admin Workstation USB to a USB port on the Qubes computer, then use the devices widget in the upper right panel to attach it to the ``vault`` VM. There will be 3 listings for the USB in the widget: one for the base USB, one for the Tails partition on the USB, labeled ``Tails``, and a 3rd unlabeled listing, for the persistent volume. Choose the third listing.
+
+- In the the ``vault`` file manager, select **+ Other Locations**., then click the persistent volume's listing in the right panel. It will be named ```N GB encrypted``, where N is the size of the persistent volume. Enter the Admin Workstation persistent volume passphrase to unlock and mount it. 
+
+- Copy the Journalist Interface configuration file to `dom0`. If your SecureDrop instance uses v3 onion services, use the following command:
```
eloquence: need double backtics on `dom0` for consistency

docs/admin/install.rst
```diff
@@ -0,0 +1,336 @@
...
+
+    securedrop-admin --validate
+
+- Finally, in the ``dom0`` terminal, run the command:
+
+  .. code-block:: sh
+
+    securedrop-admin --apply
+
+This command will take a considerable amount of time and approximately 4GB of bandwidth, as it sets up multiple VMs and installs supporting packages. When it completes, your SecureDrop Workstation is finally ready to use!
```
eloquence: The most common reason I'm aware of that this command may fail on first run are network issues in guest VMs even when the host network is up. Both Mickael and I have observed this in recent runs. In case this happens, a re-run is required. I'll open an issue to track, and if we don't resolve soon, we may want to add a note here.

docs/admin/install.rst
zenmonkeykstop: noted.

docs/admin/install.rst
zenmonkeykstop: :heavy_check_mark:

docs/admin/install.rst
zenmonkeykstop: :heavy_check_mark:

PR #22, opened 2020-03-27
Expand FAQ; updates and clarifications

eloquence:
## Status

Ready for review

## Description

- Expand FAQ with missing entries

- Normal disposable VMs do allow copy and paste, so it's not quite
  right to describe this as a general security limitation of Qubes.
  We should explain our own copy and paste limitations separately.

- The "Log out" option was removed entirely due to its session-saving
  behavior.

- Note that we do not support more than one Submission Key

### DIFF
--- a/docs/journalist/ending_session.rst
+++ b/docs/journalist/ending_session.rst
@@ -6,6 +6,6 @@ Ending your session
 When you are finished using SecureDrop Workstation, close the SecureDrop Client window and shut the computer down completely. This is to take advantage of the protections of full-disk encryption, and to avoid unauthorized access to the Workstation and the files and materials on it, which include any messages and submissions that you have downloaded.
 
 To shut down the computer, click your username in the top righthand corner of
-your screen, and select **Shut Down** from the context menu. Or, in the Qubes
-Applications menu, scroll to the bottom of the menu and select **Log Out > Shut
-Down**.
+your screen, and select **Shut Down** from the menu.
+
+.. | screenshot_shut_down |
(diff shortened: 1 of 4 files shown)


docs/journalist/ending_session.rst
```diff
@@ -6,8 +6,6 @@ Ending your session
 When you are finished using SecureDrop Workstation, close the SecureDrop Client window and shut the computer down completely. This is to take advantage of the protections of full-disk encryption, and to avoid unauthorized access to the Workstation and the files and materials on it, which include any messages and submissions that you have downloaded.
 
 To shut down the computer, click your username in the top righthand corner of
-your screen, and select **Shut Down** from the context menu. Or, in the Qubes
-Applications menu, scroll to the bottom of the menu and select **Log Out > Shut
-Down**.
+your screen, and select **Shut Down** from the menu.
```
eloquence: (Technically this is not a [context menu](https://en.wikipedia.org/wiki/Context_menu) but just a regular menu.)

docs/journalist/ending_session.rst
```diff
@@ -6,6 +6,6 @@ Ending your session
...
+
+.. | screenshot_shut_down |
```
zenmonkeykstop: The associated image is missing but that isn't related to changes in this PR.

eloquence: Ah good flag, let me push a quick fix. Ro did take a screenshot but it was of the lefthand menu we don't generally want to recommend (because it includes the session-saving "Log out" option). I'll remove the reference for now.

eloquence: Done.

PR #25, opened 2020-03-31
Add hash check for RPM post-dom0-transfer to install docs

zenmonkeykstop:
Adds a `sha256sum` check before installing the config RPM in `dom0`, to verify that it was copied successfully.


zenmonkeykstop:
Human error is one potential case  - the `qvm-run` command with this flag passes through stdout regardless, so if the user `cat`s the wrong file or no file something will still be created on the dom0 side.

eloquence:
Makes sense, thanks for the additional explanation. :)

PR #35, opened 2020-04-18
Add section about post-install config: passwords, clipboard & logs

eloquence:
## Status

Ready for review. Depends on https://github.com/freedomofpress/securedrop-workstation/pull/533, which was released with SecureDrop Workstation 0.3.0.

## Testing

I have tested the specific password manager instructions, but I do recommend re-testing those with the changes from  https://github.com/freedomofpress/securedrop-workstation/pull/533 in place.

### DIFF
--- a/docs/admin/install.rst
+++ b/docs/admin/install.rst
@@ -21,7 +21,8 @@ Install tasks:
 ~~~~~~~~~~~~~~
 
 #. Copy the submission key
-#. Copy Journalist Interface details
+#. Copy *Journalist Interface* details
+#. Copy SecureDrop login credentials
 #. Download and install SecureDrop Workstation
 #. Configure SecureDrop Workstation
 #. Test the Workstation
@@ -36,7 +37,10 @@ In order to install SecureDrop Workstation and configure it to use an existing S
 
   .. note:: A USB stick with a Type-A connector is recommended, as USB-C ports may be disabled on your computer when the BIOS settings detailed below are applied.
 
-- The SecureDrop instance's Admin Workstation and Secure Viewing Station (SVS) USBs, and the full GPG fingerprint of the submission key.
+- The SecureDrop instance's *Admin Workstation* and Secure Viewing Station (SVS) USBs, and the full GPG fingerprint of the submission key.
+- (Optional, for a single-user workstation) The *Journalist Workstation* USB for the intended user of this workstation, if you want to import their SecureDrop login credentials into the workstation's password manager.
+- The passphrases required to unlock the persistent volumes on each of these USB drives.
+
 - A working computer (Linux is recommended and assumed in this guide) to use for verification and creation of the Qubes installation medium.
 
   .. note:: A Tails USB can be used to perform the tasks below, but due to the size of the Qubes installation ISO, it may make sense to download it on another computer rather than via Tor, and then to use a USB stick to transfer it to Tails for verification and creation of the installation medium.
@@ -51,9 +55,9 @@ Pre-install tasks
 
 Verify the SecureDrop server configuration
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-In order to be used with SecureDrop Workstation, your instance must be running the latest version of SecureDrop, and the server configuration must have been updated to allow for HTTP ``DELETE`` requests. The configuration change to enable this was added in the ``0.13.0`` version of SecureDrop, released on May 29 2019. If your instance was created using this or a later version, it has the necessary changes. If not, then the ``./securedrop-admin install`` command must have been run from an Admin Workstation updated with the ``0.13.0`` code or later. To check this:
+In order to be used with SecureDrop Workstation, your instance must be running the latest version of SecureDrop, and the server configuration must have been updated to allow for HTTP ``DELETE`` requests. The configuration change to enable this was added in the ``0.13.0`` version of SecureDrop, released on May 29 2019. If your instance was created using this or a later version, it has the necessary changes. If not, then the ``./securedrop-admin install`` command must have been run from an *Admin Workstation* updated with the ``0.13.0`` code or later. To check this:
 
-- Use an Admin Workstation USB to boot into Tails, with the persistent volume unlocked and an administration password set.
+- Use an *Admin Workstation* USB to boot into Tails, with the persistent volume unlocked and an administration password set.
 - Navigate to **Applications ▸ System Tools ▸ Terminal** to open a terminal.
 - Verify that the *Journalist Interface* Apache configuration allows for HTTP ``DELETE`` using the following command:
 
@@ -70,9 +74,9 @@ In order to be used with SecureDrop Workstation, your instance must be running t
 
 - If not, then you will need to:
 
-  - Update the Admin Workstation to the current SecureDrop release version, by following the applicable upgrade guide in `our documentation <https://docs.securedrop.org>`_.
+  - Update the *Admin Workstation* to the current SecureDrop release version, by following the applicable upgrade guide in `our documentation <https://docs.securedrop.org>`_.
   - Back up the SecureDrop instance, using the `server backup <https://docs.securedrop.org/en/master/backup_and_restore.html>`_ instructions.
-  - Verify that the configuration stored on the Admin Workstation is correct by running ``cd ~/Persistent/securedrop && ./securedrop-admin sdconfig``. This command will display each setting in turn - to accept without changing, press **Enter** for each.
+  - Verify that the configuration stored on the *Admin Workstation* is correct by running ``cd ~/Persistent/securedrop && ./securedrop-admin sdconfig``. This command will display each setting in turn - to accept without changing, press **Enter** for each.
   - Update the instance configuration by running ``./securedrop-admin install``.
 
 - When the instance configuration is up to date, continue with the SecureDrop Workstation installation.
@@ -203,16 +207,20 @@ In order to decrypt submissions, your SecureDrop Workstation will need a copy of
 - In the ``vault`` file manager, select **+ Other Locations** and eject the TailsData volume, then disconnect the SVS USB.
 
 
-Copy Journalist Interface details
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. _copy_journalist:
 
-SecureDrop Workstation connects to your SecureDrop instance's API via the Journalist Interface. In order to do so, it will need the Journalist Interface address and authentication info. As the clipboard from another VM cannot be copied into ``dom0`` directly, follow these steps to copy the file into place:
+Copy *Journalist Interface* details
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-- Connect the Admin Workstation USB to a USB port on the Qubes computer, then use the devices widget in the upper right panel to attach it to the ``vault`` VM. There will be 3 listings for the USB in the widget: one for the base USB, one for the Tails partition on the USB, labeled ``Tails``, and a 3rd unlabeled listing, for the persistent volume. Choose the third listing.
+SecureDrop Workstation connects to your SecureDrop instance's API via the *Journalist Interface*. In order to do so, it will need the *Journalist Interface* address and authentication info. As the clipboard from another VM cannot be copied into ``dom0`` directly, follow these steps to copy the file into place:
 
-- In the the ``vault`` file manager, select **+ Other Locations**, then click the persistent volume's listing in the right panel. It will be named ```N GB encrypted``, where N is the size of the persistent volume. Enter the Admin Workstation persistent volume passphrase to unlock and mount it.
+- Locate an *Admin Workstation* or *Journalist Workstation* USB drive. Both hold the address and authentication info for the *Journalist Interface*; if you also want to copy the journalist user's password database, use the *Journalist Workstation* USB drive.
 
-- Copy the Journalist Interface configuration file to ``dom0``. If your SecureDrop instance uses v3 onion services, use the following command:
+- Connect the USB drive to a USB port on the Qubes computer, then use the devices widget in the upper right panel to attach it to the ``vault`` VM. There will be 3 listings for the USB in the widget: one for the base USB, one for the Tails partition on the USB, labeled ``Tails``, and a 3rd unlabeled listing, for the persistent volume. Choose the third listing.
+
+- In the the ``vault`` file manager, select **+ Other Locations**, then click the persistent volume's listing in the right panel. It will be named ```N GB encrypted``, where N is the size of the persistent volume. Enter the persistent volume passphrase to unlock and mount it.
+
+- Copy the *Journalist Interface* configuration file to ``dom0``. If your SecureDrop instance uses v3 onion services, use the following command:
 
   .. code-block:: sh
 
@@ -230,8 +238,43 @@ SecureDrop Workstation connects to your SecureDrop instance's API via the Journa
 
 - Verify that the ``/tmp/journalist.txt`` file on ``dom0`` contains valid configuration information using the command ``cat /tmp/journalist.txt`` in the ``dom0`` terminal.
 
-- In the ``vault`` file manager, select **+ Other Locations** and eject the TailsData volume, then disconnect the Admin Workstation USB. Shut down the ``vault`` VM using the Qube widget in the upper right panel.
+- If you used an *Admin Workstation* USB drive, or you don't intend to copy a password database to this workstation, safely disconnect the USB drive now. In the ``vault`` file manager, select **+ Other Locations** and eject the TailsData volume, then disconnect the USB drive.
 
+Copy SecureDrop login credentials
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Users of SecureDrop Workstation must enter their username, passphrase and two-factor code to connect with the SecureDrop server. You can manage these passphrases using the KeePassXC password manager in the ``vault`` VM. If this laptop will be used by more than one journalist, we recommend that you shut down the ``vault`` VM now (using the Qube widget in the upper right panel), skip this section, and use a smartphone password manager instead.
+
+In order to set up KeePassXC for easy use:
+
+- Add KeePassXC to the application menu by selecting it from the list of available apps in **Q > Domain: vault > vault: Qube Settings > Applications** and pressing the button labeled **>** (do not press the button labeled **>>**, which will add *all* applications to the menu).
+
+- Launch KeePassXC via **Q > Domain: vault > vault: KeePassXC**. When prompted to enable automatic updates, decline. ``vault`` is networkless, so the built-in update check will fail; the app will be updated through system updates instead.
+
+- Close the application.
+
+.. important::
+
+   The *Admin Workstation* password database contains sensitive credentials not required by journalist users. Make sure to copy the credentials from the *Journalist Workstation* USB.
+
+In order to copy a journalist's login credentials:
+
+- If a *Journalist Workstation* USB is not currently attached, connect it, attach it to the ``vault`` VM, open it in the file manager, and enter its encryption passphrase.
+
+- Locate the password database. It should be in the ``Persistent`` directory, and will typically be named ``keepassx.kdbx`` or similar.
+
+- Open a second ``vault`` file manager window (``Ctrl + N`` in the current window) and navigate to the **Home** directory.
+
+- Drag and drop the password database to copy it.
+
+- In the ``vault`` file manager, select **+ Other Locations** and eject the TailsData volume, then disconnect the *Journalist Workstation* USB. Close this file manager window.
+
+- In the file manager window that displays the home directory, open the copy you made of the password database by double-clicking it.
+
+- If the database is passwordless, KeePassXC may display a security warning when opening it. To preserve convenient passwordless access, you can protect the database using a key file, via **Database > Database settings > Security > Add additional protection > Add Key File > Generate**. This key file has to be selected when you open the database, but KeePassXC will remember the last selection.
+
+- Inspect each section of the password database to ensure that it contains only the information required by the journalist user to log in.
+
+- Close the application window and shut down the ``vault`` VM (using the Qube widget in the upper right panel).
 
 Download and install SecureDrop Workstation
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -327,7 +370,7 @@ With the key and configuration available in ``dom0``, you're ready to set up Sec
 Configure SecureDrop Workstation (estimated wait time: 60-90 minutes)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Before setting up the set of VMs used by SecureDrop Workstation, you must configure the Journalist Interface connection and submission key.
+Before setting up the set of VMs used by SecureDrop Workstation, you must configure the *Journalist Interface* connection and submission key.
 
 - To add the submission key, run the following command in the ``dom0`` terminal:
 
@@ -353,8 +396,8 @@ Before setting up the set of VMs used by SecureDrop Workstation, you must config
 - The ``config.json`` file must be updated with the correct values for your instance. Open it with root privileges in a text editor such as ``vi`` or ``nano`` and update the following fields' values:
 
   - **submission_key_fpr**: use the value of the submission key fingerprint as displayed above
-  - **hidserv.hostname**: use the hostname of the Journalist Interface, including the ``.onion`` TLD
-  - **hidserv.key**: use the value of the v2 HidServAuth token for the Journalist Interface, or the v3 private authorization key value if your SecureDrop instance uses v3 onion services
+  - **hidserv.hostname**: use the hostname of the *Journalist Interface*, including the ``.onion`` TLD
+  - **hidserv.key**: use the value of the v2 HidServAuth token for the *Journalist Interface*, or the v3 private authorization key value if your SecureDrop instance uses v3 onion services
   - **environment**: use the value ``prod``
 
 .. note::
@@ -389,9 +432,33 @@ This command will take a considerable amount of time and approximately 4GB of ba
 Test the Workstation
 ~~~~~~~~~~~~~~~~~~~~
 
-To start the SecureDrop Workstation client, double-click the SecureDrop desktop icon that was set up by the previous command. The preflight updater will start and check for updates. The system should be up-to-date and no updates should be required, but if updates are available follow the instructions in the preflight updater to apply them.
+To start the SecureDrop Client, double-click the SecureDrop desktop icon that was set up by the previous command. The preflight updater will start and check for updates. The system should be up-to-date and no updates should be required, but if updates are available follow the instructions in the preflight updater to apply them.
 
-Once the update check is complete, the client will launch. Log in using an existing journalist account and verify that sources are listed and submissions can be downloaded, decrypted, and viewed.
+Once the update check is complete, the SecureDrop Client will launch. Log in using an existing journalist account and verify that sources are listed and submissions can be downloaded, decrypted, and viewed.
+
+.. _Password Management Section:
+
+Enable password copy and paste
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If you use KeePassXC in the ``vault`` VM to manage login credentials, you can enable the user to copy passwords to the SecureDrop Client using inter-VM copy and paste. While this is relatively safe, we recommend reviewing the section :doc:`Managing Clipboard Access <managing_clipboard>` of this guide, which goes into further detail on the security considerations for inter-VM copy and paste.
+
+The password manager runs in the networkless ``vault`` VM, and the SecureDrop Client runs in the ``sd-app`` VM. To permit this one-directional clipboard use, issue the following command in ``dom0``:
+
+.. code-block:: sh
+
+   qvm-tags vault add sd-send-app-clipboard
+
+Confirm that the tag was correctly applied using the ``ls`` subcommand:
+
+.. code-block:: sh
+
+   qvm-tags vault ls
+
+To revoke this configuration change later or correct a typo, you can use the ``del`` subcommand, e.g.:
+
+.. code-block:: sh
+
+   qvm-tags vault del sd-send-app-clipboard
 
 Troubleshooting installation errors
 -----------------------------------
--- a/docs/admin/managing_clipboard.rst
+++ b/docs/admin/managing_clipboard.rst
@@ -1,0 +1,71 @@
+Managing Clipboard Access
+=========================
+
+.. include:: ../includes/top-warning.rst
+
+Every VM in Qubes has its own clipboard, similar to the clipboard of a Mac, Windows or Linux computer. For example, if you used the default ``work`` VM to browse the web and wanted to copy text from one browser window to another, you would use the ``Ctrl+C`` and ``Ctrl+V`` keyboard shortcuts to copy and paste. This type of clipboard usage -- copy and paste in the same VM -- also works in all VMs that are part of SecureDrop Workstation.
+
+In addition, Qubes supports copying information *between* VMs. This is done by using `special keyboard shortcuts <https://www.qubes-os.org/doc/copy-paste/>`_, ``Ctrl+Shift+C`` and ``Ctrl+Shift+V``, in a four-step process. By default, this is disabled for all VMs that are part of SecureDrop Workstation, consistent with the `principle of least privilege <https://en.wikipedia.org/wiki/Principle_of_least_privilege>`__.
+
+As an administrator, you should be aware of the following risks related to clipboard access before changing the default configuration:
+
+1. It is dangerous to copy untrusted, unsanitized content *into* a secure environment. What looks like plain text may contain character sequences that exploit security vulnerabilities in the target environment.
+2. The four-step process described above can be difficult to follow, and it is easy to make an operational mistake, such as pasting a password into a message to a source, or into a window belonging to a VM with network access.
+3. Like any other part of the operating system, the implementation of Qubes clipboard itself may contain undiscovered security vulnerabilities that an adversary could exploit in an attempt to exfiltrate information.
+
+With these considerations in mind, there are use cases where clipboard access may be an important part of your regular use of SecureDrop Workstation. For example:
+
+- You may want to copy passwords from a password manager to the SecureDrop Client;
+- You may want to copy a message you received via SecureDrop into a secure messaging app like Signal, to share it with another journalist.
+
+To support these use cases, SecureDrop Workstation allows you to grant granular access to the ``sd-app`` clipboard (via the cross-VM clipboard) to selected VMs.
+
+Configuring clipboard access to ``sd-app``
+------------------------------------------
+
+The process for permitting the one-directional copying of passwords from a password manager in ``vault`` to the SecureDrop Client is :ref:`outlined in the installation docs <Password Management Section>`. In general, clipboard access to SecureDrop Workstation VMs is governed by *tags* that can be applied in ``dom0`` to selected VMs:
+
+- the tag ``sd-send-app-clipboard`` can be used to tag a VM that should be able to send its clipboard contents *to* ``sd-app`` via the cross-VM clipboard;
+- the tag ``sd-receive-app-clipboard`` can be used to tag a VM that should be able to receive its clipboard contents *from* ``sd-app`` via the cross-VM clipboard.
+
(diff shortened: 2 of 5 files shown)


eloquence:
(CI failure is a linkcheck error unrelated to this PR; may be a flake as the reported link  https://docs.securedrop.org/en/stable/journalist.html#flag-for-reply  works fine for me.)

docs/admin/install.rst
```diff
@@ -336,9 +336,77 @@ This command will take a considerable amount of time and approximately 4GB of ba
...
+
+  qvm-tags vault del sd-send-app-clipboard
+
+These policy changes take effect immediately.
+
+Reviewing and exporting logs
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+SecureDrop Workstation aggregates system logs from all its VMs in the ``sd-log`` VM, in the folder ``~/QubesIncomingLogs``, with one subfolder for each VM. While details about submissions are not logged, the logs may contain sensitive information, such as timing data, the two-word designation for a given source, or revelatory errors. For this reason, the ``sd-log`` VM is networkless, and you cannot copy files from ``sd-log`` to other VMs by default.
+
+If you want to selectively enable copying logs to a target VM such as ``work``, you can use a command like the following in ``dom0``:
```
emkll, line 393: There should be also be a disclaimer here about the risks of sharing these logs with other VMs, specifically that logs contain Submission and reply metadata, as well as usage information pertaining to the SecureDrop workstation and servers that is not public. It might make more sense to have all these commands consolidated later, where the risks are clearly defined.

docs/admin/managing_clipboard.rst
```diff
@@ -0,0 +1,53 @@
+Managing Clipboard Access
+=========================
+
+.. include:: ../includes/top-warning.rst
+
+Every VM in Qubes has its own clipboard, similar to the clipboard of a Mac, Windows or Linux computer. For example, if you wanted to create a boilerplate "Thank you" message for replies to sources, you could create a textfile in the ``sd-app`` VM and copy its contents to the SecureDrop Client using ``Ctrl+C`` (copy) and ``Ctrl+V`` (paste) keyboard shortcuts.
+
+Qubes also supports copying information *between* VMs. This is done by using `special keyboard shortcuts <https://www.qubes-os.org/doc/copy-paste/>`_, ``Ctrl+Shift+C`` and ``Ctrl+Shift+V``, in a four-step process. By default, this is disabled for all VMs that are part of SecureDrop Workstation. There are three primary reasons for this:
+
+1. It is dangerous to copy untrusted content *into* a secure environment. What looks like plain text may contain character sequences that exploit security vulnerabilities in the target environment.
```
emkll, line 10: Qualifying copying untrusted content into secure environment with `without proper sanitization` might be appropriate here, since this is effectively what the client does. To me (based on https://github.com/freedomofpress/securedrop-workstation/pull/533#pullrequestreview-398264882), the concept of integrity is useful here, but defer to what the team thinks is best to convey these concepts to the reader.

docs/admin/managing_clipboard.rst
```diff
@@ -0,0 +1,53 @@
...
+2. The four-step process described above can be difficult to follow, and it is easy to make an operational mistake, such as pasting the wrong content into a message to a source.
```
emkll, line 11: or into another VM or other window, affecting the confidentiality of messages

docs/admin/managing_clipboard.rst
```diff
@@ -0,0 +1,53 @@
...
+3. SecureDrop Workstation aims to minimize the `attack surface <https://en.wikipedia.org/wiki/Attack_surface>`__ adversaries can target to exfiltrate information. For example, if the clipboard implementation in Qubes OS itself contained a security vulnerability, an adversary could attempt to exploit that vulnerability to shuttle a message sent by a source from the networkless ``sd-app`` VM to a networked VM.
```
emkll, line 12: Reduction of attack surface is an overarching theme, and is also design rationale for point 1. in this list. There is another risk that is not elaborated here, which is the principle of least privilege. You want to ensure that all VMs have access to the functionality that they need, and that this functionality shouldn't be enabled if it's not required: Should sd-app be compromised, it could potentially be easier to exfiltrate data from the workstation

docs/admin/install.rst
```diff
@@ -336,9 +336,77 @@ This command will take a considerable amount of time and approximately 4GB of ba
...
-Once the update check is complete, the client will launch. Log in using an existing journalist account and verify that sources are listed and submissions can be downloaded, decrypted, and viewed.
+.. code-block:: sh
+
+  qvm-tags vault add sd-send-app-clipboard
+
+To remove the permission, use:
+
+.. code-block:: sh
+
+  qvm-tags vault del sd-send-app-clipboard
```
emkll, line 385: Running `qvm-tags vault ls` after the fact would help confirm the tag has been successfully applied to the correct VM. It's very important these tags are properly applied as there is nothing in automation tooling to audit these permissions. It is an  inherently risky action

docs/admin/install.rst
```diff
@@ -336,9 +336,77 @@ This command will take a considerable amount of time and approximately 4GB of ba
...
+Qubes supports copying the clipboard contents from one VM to another using `special keyboard shortcuts <https://www.qubes-os.org/doc/copy-paste/>`_. It also permits configuring the access policies for each VM's clipboard. By default, cross-VM clipboard copying is completely disabled for all VMs created as part of the installation (the ones that start with ``sd-``) for security reasons.
+
+..  important::
+
+    Before changing the clipboard configuration, we recommend reviewing the section :doc:`Managing Clipboard Access <managing_clipboard>` of this guide, which goes into further detail on the security considerations.
+
+If you did set up a password manager in the previous step, you may want to enable copying passwords from KeePassXC to the SecureDrop Client. The password manager runs in the networkless ``vault`` VM, and the SecureDrop Client runs in the ``sd-app`` VM. To permit this one-directional clipboard use, issue the following command in ``dom0``:
 
-Once the update check is complete, the client will launch. Log in using an existing journalist account and verify that sources are listed and submissions can be downloaded, decrypted, and viewed.
+.. code-block:: sh
```
emkll, line 377: There's a lot of overlap in the commands in the `Managing clipboard access` section, and it might make sense to consolidate the examples (for example move this one to the latter section and link to it). The language used here combined with the linking to information found later in the document will make it unlikely that a user will actually read the details prior to setting up clipboard sharing

docs/admin/install.rst
```diff
@@ -336,9 +336,77 @@ This command will take a considerable amount of time and approximately 4GB of ba
...
+
+The ``fedora-30`` template includes the KeePassXC password manager. It is not listed in the application menu for ``vault`` by default; you can add it by selecting it from the list of available apps in **Q > Domain: vault > Qube Settings > Applications** and pressing the button labeled **>** (do not press the button labeled **>>**, which will add *all* applications to the menu).
+
+After confirming this change, you can launch KeePassXC from the **Domain: vault** menu. On first start, it will suggest enabling automatic updates, which you should decline: ``vault`` is networkless, so the built-in update check will fail; the app will be updated through system updates instead.
+
+When creating a new password database, KeePassXC will prompt you to protect it with a password, and will refuse to accept a blank password. If you want to configure easy access to the password database, you can protect the database using a key file, via **Add additional protection > Add Key File > Generate** on the final screen. Once you have configured a key file, KeePassXC will accept leaving the master password field blank.
```
emkll, line 365: Do we provide any instructions on how to copy the password itself to this new database?

docs/admin/managing_clipboard.rst
```diff
@@ -0,0 +1,53 @@
...
+
+With these considerations in mind, there are use cases where clipboard access may be an important part of your regular use of SecureDrop Workstation. For example:
+
+- You may want to copy passwords from a password manager to the SecureDrop Client;
+- You may want to copy a message you received via SecureDrop into a secure messaging app like Signal, to share it with another journalist.
+
+To support these use cases, SecureDrop Workstation allows you to grant granular access to the ``sd-app`` clipboard (via the cross-VM clipboard) to selected VMs.
```
emkll, line 19: By granular do you mean read or write or on a per-vm basis? Or all three?

docs/admin/managing_clipboard.rst
```diff
@@ -0,0 +1,53 @@
...
+
+Configuring clipboard access to ``sd-app``
```
emkll, line 21: you refer to `signal` vm here which might lead to confusion, as there are currently no signal VMs provisioned. It might  make sense to state that we should expect, as a general rule, that these rules should be applied (for now) to _user managed_ VMs

docs/admin/managing_clipboard.rst
```diff
@@ -0,0 +1,53 @@
...
+
+The syntax for revoking a tag is as follows:
+
+.. code-block:: sh
+
+  qvm-tags <VM name> del <tag name>
+
+Changes to tags take effect immediately, and any VM can have multiple tags.
+
+.. important::
```
emkll, line 49: I would suggest we move this at the top of this page

docs/admin/install.rst
eloquence: It currently says:

> the logs may contain sensitive information, such as timing data, the two-word designation for a given source, or revelatory errors.

I can spell that out in a bit more detail as a warning.

docs/admin/managing_clipboard.rst
eloquence: I think rewriting that point 3 under "principle of least privilege" makes sense.

docs/admin/managing_clipboard.rst
eloquence: :+1:

docs/admin/install.rst
eloquence: Not currently. May want to test importing an Admin Workstation KeePassXC DB into Qubes ...

docs/admin/managing_clipboard.rst
eloquence: All three.

docs/admin/install.rst
eloquence: Added further instructions in cf923f620bbff1eb63cdb3df060356a6b6ea0ba2.

docs/admin/install.rst
eloquence: Done in 5d221f0eca63259c01d8583c059930667ea0bddd. I've split this into its own admin section in this commit; in spite of the command similarities with clipboard access, my bias is to keep it in a separate section, since we may want to expand each section in different ways (e.g., we may want to flesh out the logs section in future to give some more pointers about how to redact logs, what to look for, etc.).

docs/admin/managing_clipboard.rst
eloquence: Added "unsanitized" in 429dbb26a2fc82305455b5f4b927ac8f615c203c

docs/admin/managing_clipboard.rst
eloquence: Added similar language in 829a0f74a79ec52c88ea1a6e6fe2b9fc71397bc7 (wanted a somewhat more concrete example)

docs/admin/managing_clipboard.rst
eloquence: Upon consideration, used that language the intro, and tried to make the bullets a bit more concrete, see 829a0f74a79ec52c88ea1a6e6fe2b9fc71397bc7

docs/admin/install.rst
eloquence: Good point, added a suggestion to use ``ls`` subcommand, throughout all uses of this tooling.

docs/admin/install.rst
eloquence: I would prefer not to decouple the concrete instructions for setting up password copying from the setup of the password manager -- those two really belong together, as most users will want both to make meaningful use of the password manager. Additionally, the risks of this single pathway for clipboard copying (one way, from ``vault`` to ``sd-app``) seem among the most manageable, especially if we land the password-wipe logic in the client along with this change.

docs/admin/managing_clipboard.rst
eloquence: I think it's generally a good practice in documentation to set examples apart a bit, I've tried to do that in 829a0f74a79ec52c88ea1a6e6fe2b9fc71397bc7.

docs/admin/managing_clipboard.rst
eloquence: Moved up, but not before the terminology is introduced.

docs/admin/install.rst
eloquence: Oh, also, because this is in the post-install config, the USB device will be grabbed by `sd-devices` first. Will make a note of that.

docs/journalist/faq.rst
```diff
@@ -139,16 +139,17 @@ without opening it in an interactive viewer application.
 
 Why can't I copy and paste?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
-In the current version, copy and paste between and to SecureDrop Workstation VMs
-is disabled for security reasons. The goal of this restriction is to minimize
-the risk of accidental pastes of sensitive content, and to reduce the attack
-surface for attempts to exfiltrate information.
+You should be able to copy and paste *within* any VM on the system, e.g.,
+from one application running in ``sd-app`` to another.
```
rocodes, line 143: this is a nit, but I wonder if we should pick a different example vm to discourage running any other app than the SecureDrop client in `sd-app`?

rocodes: Ah, and just a wee merge conflict due to #36 being merged.

rocodes:
And a wee merge conflict to resolve due #36 being merged

docs/journalist/faq.rst
eloquence: I'm fine with changing this example to copying text within `work` (a default Qubes VM) to sidestep the question. I do think we should talk a bit more about such use cases as they arise -- e.g., the boilerplate use case. I'm not sure it really _is_ better to recommend using a separate VM in a case like that, because then the user has to punch in RPC holes to allow cross-VM copy/paste, and manage the copy/paste opsec every time -- plus the VM may well have worse security properties than `sd-app`, if we don't ship it. But this is probably not something we want to speak to in a casual example.

eloquence:
As a general note, we don't have a `stable` or `prod` branch for our docs live build yet, so the moment this gets merged, it'll be the default view in https://workstation.securedrop.org/ . Since these changes aren't in any prod release yet, I am wondering if we should
- hold off on merge until then, or
- revisit the branch structure for this repo

docs/admin/install.rst
eloquence: Clarified `sd-devices` behavior in b07ac1948fe4c804320087525e85901fb183f269.

docs/admin/reviewing_logs.rst
```diff
@@ -0,0 +1,36 @@
...
+- error messages that disclose further details
+
+For this reason, the ``sd-log`` VM is networkless, and you cannot copy files from ``sd-log`` to other VMs by default.
+
+If you want to selectively enable copying logs to a single VM, you can use tags, similar to the method used for :doc:`managing clipboard access <managing_clipboard>`. You can add and remove the permission just before each copying operation; the change will take effect immediately.
+
+.. important::
+
+   Before copying logs to a networked VM, we recommend carefully inspecting them for sensitive information, and potentially redacting them
+
```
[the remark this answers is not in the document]
eloquence: Fixed in b07ac1948fe4c804320087525e85901fb183f269

docs/admin/reviewing_logs.rst
```diff
@@ -0,0 +1,36 @@
...
+  qvm-tags <VM name> ls
+
+To remove the permission, use this command in ``dom0``:
+
+.. code-block:: sh
+
+  qvm-tags <VM name> del sd-receive-logs
+
+With the permission in effect, you can use the command ``qvm-copy`` in a terminal in ``sd-log`` to copy individual files to the target VM.
```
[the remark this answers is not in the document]
eloquence: Why not both? :) Done in b07ac1948fe4c804320087525e85901fb183f269

docs/journalist/faq.rst
eloquence: Changed to a different example in b07ac1948fe4c804320087525e85901fb183f269

eloquence:
Attempted to address all comments and also squashed & rebased onto latest `master`, with most recent changes still in a separate commit for easier review. I'd appreciate if you could take a final spin @rocodes, but would still recommend holding off on merge until we have team-level agreement on how we want to handle docs changes that aren't applicable to our current production release.

rocodes:
Thank you for those changes--LGTM!

eloquence:
Rebased.

eloquence:
(Rebased to latest `master`.)

docs/admin/install.rst
```diff
@@ -389,9 +391,77 @@ This command will take a considerable amount of time and approximately 4GB of ba
...
-To start the SecureDrop Workstation client, double-click the SecureDrop desktop icon that was set up by the previous command. The preflight updater will start and check for updates. The system should be up-to-date and no updates should be required, but if updates are available follow the instructions in the preflight updater to apply them.
+To start the SecureDrop Client, double-click the SecureDrop desktop icon that was set up by the previous command. The preflight updater will start and check for updates. The system should be up-to-date and no updates should be required, but if updates are available follow the instructions in the preflight updater to apply them.
+
+Once the update check is complete, the SecureDrop Client will launch. Log in using an existing journalist account and verify that sources are listed and submissions can be downloaded, decrypted, and viewed.
+
+Post-Install Configuration
+--------------------------
+Once you have verified that you can log into the SecureDrop Client, you may want to consider making additional configuration changes:
+
+- setting up password management
```
zenmonkeykstop, line 402: A workstation USB is attached during the installation phase, it would make sense to copy the keepassx db across at that point (as an optional step).
- this avoids repeatedly connecting said USB during the course of the install,
- it avoids fun with USB auto-attach and associated potential issues (thinking in particular of the one where encrypted volumes don't seem to get removed properly and create multiple `/dev/xvdiN` devices.),
- the docs don't have to explain the `vault` VM twice.

docs/admin/install.rst
```diff
@@ -389,9 +391,77 @@ This command will take a considerable amount of time and approximately 4GB of ba
...
+
+.. _Password Management Section:
+
+Password management
+~~~~~~~~~~~~~~~~~~~
+If you have installed Qubes with the default configuration, you should have a VM called ``vault``, which is a networkless VM based on the ``fedora-31`` template. You can use this VM to store SecureDrop login credentials or the passphrases for encrypted export USB drives.
+
+Note that all users of this workstation will have access to the password database. If the workstation will be used by more than one journalist, you may want to encourage the use of a smartphone password manager for managing individual login credentials.
+
+The ``fedora-31`` template includes the KeePassXC password manager. It is not listed in the application menu for ``vault`` by default; you can add it by selecting it from the list of available apps in **Q > Domain: vault > Qube Settings > Applications** and pressing the button labeled **>** (do not press the button labeled **>>**, which will add *all* applications to the menu).
```
zenmonkeykstop, line 416: See also point above, could be reworded as "The KeePassXC password manager is available in the `vault` VM, but is not listed in the application menu by default; [...]"

docs/admin/install.rst
```diff
@@ -389,9 +391,77 @@ This command will take a considerable amount of time and approximately 4GB of ba
...
+
+After confirming this change, you can launch KeePassXC from the **Domain: vault** menu. On first start, it will suggest enabling automatic updates, which you should decline: ``vault`` is networkless, so the built-in update check will fail; the app will be updated through system updates instead.
+
+You can now choose between creating a new password database or importing an existing one.
```
zenmonkeykstop, line 420: If moved into the install section, no need to mention creating a new database.

docs/admin/install.rst
```diff
@@ -389,9 +391,77 @@ This command will take a considerable amount of time and approximately 4GB of ba
...
+
+..  important::
+
+    If you import an existing database, make sure you import only the passwords you need. The password database stored on the *Admin Workstation* USB is intended to hold credentials required for the *administration* of the SecureDrop servers, which journalist users should not have access to. In contrast, a *Journalist Workstation* USB is intended to only hold the SecureDrop login credentials for a given journalist user of the system.
+
+    We recommend reviewing each section of the password database before giving other users access to the workstation.
```
zenmonkeykstop, line 426: This is more an overall style point - the use of "we" was avoided in the rest of this doc, it might be worthwhile being consistent with that. (Who is this mysterious "we" anyway? They never introduce themselves, that's a little creepy.)

docs/admin/install.rst
```diff
@@ -389,9 +391,77 @@ This command will take a considerable amount of time and approximately 4GB of ba
...
+
+To import an existing database, attach the required USB drive (typically a *Journalist Workstation* USB) to the ``vault`` VM, similar to the process described in :ref:`copy_journalist`. Note that you may have to detach the USB drive from the ``sd-devices`` VM, first, which will attempt to automatically attach USB storage devices and printers for use by SecureDrop Workstation.
```
zenmonkeykstop, line 428: This and subsequent paras are part of a sequential process and could be list-formatted.

docs/admin/install.rst
```diff
@@ -389,9 +391,77 @@ This command will take a considerable amount of time and approximately 4GB of ba
...
+
+Clipboard configuration
+~~~~~~~~~~~~~~~~~~~~~~~
+Qubes supports copying the clipboard contents from one VM to another using `special keyboard shortcuts <https://www.qubes-os.org/doc/copy-paste/>`_. It also permits configuring the access policies for each VM's clipboard. By default, cross-VM clipboard copying is completely disabled for all VMs created as part of the installation (the ones that start with ``sd-``) for security reasons.
+
+..  important::
+
+    Before changing the clipboard configuration, we recommend reviewing the section :doc:`Managing Clipboard Access <managing_clipboard>` of this guide, which goes into further detail on the security considerations.
+
+If you did set up a password manager in the previous step, you may want to enable copying passwords from KeePassXC to the SecureDrop Client. The password manager runs in the networkless ``vault`` VM, and the SecureDrop Client runs in the ``sd-app`` VM. To permit this one-directional clipboard use, issue the following command in ``dom0``:
```
zenmonkeykstop, line 444: As above, this wording would have to change if password stuff moved up in the process.

docs/admin/install.rst
zenmonkeykstop: I'd agree with the original reviewer on this one. Folks should be made aware of the associated risks before the process to add/remove tags is presented, and since there's a dedicated page for that it would make sense to consolidate it IMO. (Also gives us a single location to update if this process changes in the future.)

docs/admin/install.rst
```diff
@@ -389,9 +391,77 @@ This command will take a considerable amount of time and approximately 4GB of ba
...
+
+.. code-block:: sh
+
+  qvm-tags vault add sd-send-app-clipboard
+
+We recommend confirming that the tag was correctly applied using the ``ls`` subcommand:
```
zenmonkeykstop, line 450: See note above re: general style and use of first person plural.

docs/admin/managing_clipboard.rst
```diff
@@ -0,0 +1,64 @@
...
+
+To support these use cases, SecureDrop Workstation allows you to grant granular access to the ``sd-app`` clipboard (via the cross-VM clipboard) to selected VMs.
+
+Configuring clipboard access to ``sd-app``
+------------------------------------------
+
+The process for setting up a password manager and permitting the one-directional copying of passwords to the SecureDrop Client is :ref:`outlined in the installation docs <Password Management Section>`. In general, clipboard access to SecureDrop Workstation VMs is governed by *tags* that can be applied in ``dom0`` to selected VMs:
+
+- the tag ``sd-send-app-clipboard`` can be used to tag a VM that should be able to send its clipboard contents *to* ``sd-app`` via the cross-VM clipboard;
+- the tag  ``sd-receive-app-clipboard`` can be used to tag a VM that should be able to receive is clipboard contents *from* ``sd-app`` via the cross-VM clipboard.
```
zenmonkeykstop, line 29: nit: s/is clipboard/its clipboard

docs/admin/managing_clipboard.rst
```diff
@@ -0,0 +1,64 @@
...
+
+  qvm-tags <VM name> ls
+
+The syntax for revoking a tag is as follows:
+
+.. code-block:: sh
+
+  qvm-tags <VM name> del <tag name>
+
+As before, we recommend confirming the operation via the ``ls`` subcommand.
```
zenmonkeykstop, line 57: It might also be worth using `qvm-ls  --tags=` as a final review step, to verify that the tags have been applied to only the expected VMs

docs/admin/reviewing_logs.rst
```diff
@@ -0,0 +1,42 @@
+Reviewing and exporting logs
```
zenmonkeykstop, line 1: This discusses log export, not review - it might make sense to consolidate it into the clipboard management page instead, giving the latter a structure like:
- risks and benefits
- example: enabling one-way copy/paste from the `vault` VM to `sd-app`
- example: enabling two-way copy/paste between `sd-app` and a user-managed VM
- example: exporting logs from `sd-logs`.

docs/admin/install.rst
eloquence: I like that suggestion a lot especially because of the second point (auto-attach confusion). In organizations where journalist and admin are not the same person, it does mean a third USB drive (Journalist Workstation USB) may hold those credentials, but that's all the more reason to add it to the section upfront where users have to inventory USBs. Will rejig.

docs/admin/install.rst
eloquence: For comparison, here are all instances of the phrase "We recommend" in the SecureDrop core docs. In my view, use of that phrase is reasonable -- even if the user does not know exactly who "We" is, it implicitly speaks to recommendations from the developers of the system. 

There are alternative phrases, but the use of passive voice ("It is recommended") that's often substituted as an alternative has its own problems; it's impersonal and begs the question "by whom".

<details>
<summary>Click to expand</summary>

```
ossec_alerts.rst
359:SMTP relay securely. We recommend doing this before running the

set_up_tails.rst
28:We recommend creating an initial Tails Live DVD or USB, and then using
95:   passphrase that's easy to write down or remember. We recommend using
105:   passphrases. We recommend using `spaced-repetition

journalist.rst
232:    in a given session. We recommend storing this passphrase in your own
445:and flac. We recommend using this and other tools to work with documents within

servers.rst
36:them to your Admin Workstation. We recommend saving them to the
257:several times during SecureDrop's installation. We recommend that the

passphrase_best_practices.rst
25:   .. tip:: We recommend reviewing your passphrase (e.g. by ensuring that you
79:user role in :doc:`passphrases`. We recommended using the KeePassXC password

https_source_interface.rst
82:   unauthorized access. We recommend doing this entire process on the Admin

hardware.rst
206:We recommend using one or multiple designated USB drives as the *Transfer
328:indistinguishable from each other unless you label them. We recommend buying
383:NUC before it can be used. We recommend:
562:persistent partition. We recommend getting something in the 16-64GB
581:We recommend the `pfSense SG-3100

admin.rst
31:manual action if requested in the `SecureDrop Release Blog`_. We recommend
357:server, their encrypted files may still exist in backups. We recommend
394:instance via the Admin Interface. We recommend a size of ``500px x 450px``.
453:We recommend admins establish a secure communication pipeline to alert fellow admins

update_bios.rst
59:Power off the *Monitor Server*. We recommend plugging it into an uninterrupted power supply (UPS). Plug in the keyboard, monitor, and USB key, and power on the server, then press F7 when prompted to enter the BIOS Update tool.

overview.rst
150:admin. We recommend that you set aside at least a week to

includes/encrypting-drives.txt
4:   encrypted and protected with a secure passphrase. We recommend using the

includes/otp-app.txt
1:.. tip:: We recommend using FreeOTP (available `for Android <https://play.google.com/store/apps/details?id=org.fedorahosted.freeotp>`__

set_up_transfer_and_export_device.rst
181:We recommend using a fully encrypted USB drive for copying files off the
190:We recommend the use of VeraCrypt. It is actively maintained cross-platform
206:   We recommend against installing the VeraCrypt software on the
222:that the drive must work across platforms (including Tails). We recommend

network_firewall.rst
279:#. Set a strong admin passphrase. We recommend generating a strong passphrase
380:	  We recommend keeping it simple and using the same DNS

source.rst
91:We recommend conducting all research related to your submission in Tor Browser.
167:   you adjust the Security Setting again. We recommend keeping the setting at

rebuild_admin.rst
432:We recommend completing the following tasks after the rebuild:

glossary.rst
119:network device. We recommend physically removing the drive and any
147:authentication codes. We recommend using one of:

update_tails_usbs.rst
29:We recommend that you :ref:`back up your existing configuration <backup_workstations>` and update as soon as possible. It usually takes some time for updates to download, so keep that in mind when choosing when to update.

development/setup_development.rst
32:We recommend using the stable version of Docker CE (Community Edition) which can
180:We recommend using the latest stable version of Vagrant, ``1.8.5`` at the time
208:provision VMs. We recommend installing Ansible from PyPi with ``pip`` to ensure
263:The version of rsync installed by default on macOS is extremely out-of-date, as is Apple's custom. We recommend using Homebrew_ to install a modern version (3.1.0 or greater): ``brew install rsync``.
```
</details>

docs/admin/reviewing_logs.rst
eloquence: There are two main issues I have with that:
- it makes the documentation less task-centric and more developer-centric
- it conflates risks and benefits of two very different RPC policy changes.

On the first point, as an admin, my understanding of the system may not extend to RPC policies. I likely have a very specific goal: enable copy and paste, get at the logs, etc. Right now it's very easy to find that specific _goal_ in the table of contents.

On the second point, the permission to copy files from a VM that only collects logs to another VM is very different in terms of its security risks from the clipboard-related permissions.

That said, let me look at the overall shape of the docs once we move stuff around in the install section (will need to test that procedure as well).

docs/admin/install.rst
eloquence: Discussed that a bit more, I agree it makes sense to get rid of some of the "we recommend"s that are specifically in the context of step-by-step instructions where it can be distracting and suggests more choice than is warranted.

docs/admin/install.rst
eloquence: This is now done in dd9242951e399f860eae68bf2ed42b75cf0d11ad and I feel it definitely streamlines the process, but please take a close look at those revised instructions as this is a big change and new issues may have crept in.

docs/admin/install.rst
eloquence: Ultimately this felt like extraneous detail to begin with so I've kept it more task-focused in dd9242951e399f860eae68bf2ed42b75cf0d11ad.

docs/admin/install.rst
eloquence: Agreed, no longer mentioned in dd9242951e399f860eae68bf2ed42b75cf0d11ad

docs/admin/install.rst
eloquence: Done in dd9242951e399f860eae68bf2ed42b75cf0d11ad and f28f25ed3f46229b5b55b842c46b644065912622

docs/admin/install.rst
eloquence: Done in dd9242951e399f860eae68bf2ed42b75cf0d11ad

docs/admin/managing_clipboard.rst
eloquence: Fixed.

docs/admin/managing_clipboard.rst
eloquence: Yup (syntax printed out by the command has no `=` and that only works for a single tag argument, so I've omitted it).

docs/admin/install.rst
eloquence: To keep a task-focused installation guide in a single section of the docs, what I've done in dd9242951e399f860eae68bf2ed42b75cf0d11ad is reduce the instructions for setting up password copying (which is one of the safer operations, and almost certainly desirable if you're using a password manager) to a minimum. If you strongly feel that it should not be in the install guide, feel free to push a commit to consolidate it, and I can review.

docs/admin/reviewing_logs.rst
eloquence: I still prefer the current organization for these reasons, but as with the password management, if you feel strongly differently, feel free to push a commit to reorganize and I can review. I would just ask that we try to make it as easy as possible for an admin to locate the specific instructions they need to complete the task they're interested in.

docs/admin/install.rst
zenmonkeykstop: Given time constraints, will think about this some more and maybe do a follow-up PR.

docs/admin/install.rst
zenmonkeykstop: as it's moved to initial install, the sd-devices stuff is moot :)

docs/admin/reviewing_logs.rst
zenmonkeykstop: Exfiltration risk via logs or clipboard are pretty similar; but again, due to release timing this can go as-is with a potential follow-up PR.

docs/admin/install.rst
```diff
@@ -37,6 +38,9 @@ In order to install SecureDrop Workstation and configure it to use an existing S
   .. note:: A USB stick with a Type-A connector is recommended, as USB-C ports may be disabled on your computer when the BIOS settings detailed below are applied.
 
 - The SecureDrop instance's Admin Workstation and Secure Viewing Station (SVS) USBs, and the full GPG fingerprint of the submission key.
+- The Journalist Workstation USB for the intended user of this workstation, if you intend to use the password manager on SecureDrop Workstation.
```
zenmonkeykstop, line 41: This could probably be an either-or situation, a JW is going to have the necessary credentials for the journalist interface. So something like (but with better wording):

> - A workstation USB:
>   - if the SDW is being set up for a single journalist, use their JW to copy the necessary credentials (JI and optionally passwords)
 >   - if the SDW is being set up for use by multiple jourrnalists, use a JW or admin workstation to copy over the JI details only

eloquence: Yup that makes sense, will take a stab after standup.

docs/admin/install.rst
```diff
@@ -230,8 +236,43 @@ SecureDrop Workstation connects to your SecureDrop instance's API via the Journa
...
+
+Copy SecureDrop login credentials
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Users of SecureDrop Workstation must enter their username, passphrase and two-factor code to connect with the SecureDrop server. You can manage these passphrases using the KeePassXC password manager in the ``vault`` VM. If this laptop will be used by more than one journalist, we recommend that you shut down the ``vault`` VM now (using the Qube widget in the upper right panel), skip this section, and use a smartphone password manager instead.
+
+In order to set up KeePassXC for easy use:
+
+- Add KeePassXC to the application menu by selecting it from the list of available apps in **Q > Domain: vault > Qube Settings > Applications** and pressing the button labeled **>** (do not press the button labeled **>>**, which will add *all* applications to the menu).
+
+- Launch KeePassXC from the **Domain: vault** menu. When prompted to enable automatic updates, decline. ``vault`` is networkless, so the built-in update check will fail; the app will be updated through system updates instead.
```
zenmonkeykstop, line 249: For consistentcy, should provide full path.

docs/admin/install.rst
```diff
@@ -230,8 +236,43 @@ SecureDrop Workstation connects to your SecureDrop instance's API via the Journa
...
+
+- Connect the *Journalist Workstation* USB, attach it to the ``vault`` VM, and open it in the file manager, and enter the passphrase for this specific *Journalist Workstation* USB drive.
+
+- Locate the password database. It should be in the ``Persistent`` directory, and will typically be named ``keepassx.kdbx`` or similar.
+
+- Open a second ``vault`` file manager window (``Ctrl + N`` in the current window) and navigate to the **Home** directory.
+
+- Drag and drop the password database to copy it.
+
+- In the ``vault`` file manager, select **+ Other Locations** and eject the TailsData volume, then disconnect the *Journalist Workstation* USB. Shut down the ``vault`` VM using the Qube widget in the upper right panel.
```
zenmonkeykstop, line 267: This step asks the user to shut down the vault vm, but the next step asks them to click something in the vault VM file manager window. Probably need to remove the shutdown, or add a task to start it up again.

docs/admin/install.rst
eloquence: Upon consideration, we need the Admin Workstation USB during the SecureDrop install verification stage, so it has to be in the inventory no matter what. I'm leaving the existing wording there, but will make the section about copying over the JI details a bit more flexible in terms of which USB is used.

docs/admin/install.rst
eloquence: The next step is referring to opening the copy they made in the home directory. I've added the vault shutdown before this step, to ensure they're not making any unintended writes to the copy on the Tails USB drive. I'll make the references to the respective windows a bit clearer.

zenmonkeykstop: Addressed in subsequent commit: :heavy_check_mark:

eloquence:
(Squashed into one commit prior to merge.)

PR #43, opened 2020-05-20
added notes to support use of Tails during the installation

### LINKED ISSUE #28: For install second laptop, recommend Tails as fallback if no Linux laptop available
For near future SD Workstation installs, we can assume that they have existing Admin Workstation USBs. Where we currently instruct users to perform operations on a second Linux laptop, we should recommend booting a laptop with a Tails USB (keeping any persistent volume locked) as a fallback.

Note that some websites that may be loaded in the course of the preparatory steps will require use of the unsafe browser in Tails (especially the Lenovo website for T480 updates), due to Tor being blocked.

The alternative is to this process is to test and maintain procedures for Mac/Windows, which we don't have the bandwidth to do.

zenmonkeykstop:
Closes #28 

Adds useful info for users choosing Tails as their working computer for the install process.

## Testing
- [ ] does CI pass?
- [ ] are the docs clear and correct?

### DIFF
--- a/docs/admin/hardware.rst
+++ b/docs/admin/hardware.rst
@@ -37,6 +37,28 @@ The instructions below assume the use of a Linux-based computer for the creation
 - Visit `<https://support.lenovo.com>`_ in the Linux-based computer. Type the machine type found above into the search bar, then press **Enter**.
 - In the T480 Product Home page, select **Drivers And Software** and choose **BIOS/UEFI**.
 - Expand the **BIOS Update** listing and download the **BIOS Update (Bootable CD)** file.
+
+.. note::
+  A Tails USB can be used for the verification and conversion process described below, but the Lenovo support site blocks requests over Tor, preventing the ISO download. To work around this, either:
+
+  - download the BIOS ISO on a different computer and transfer it to Tails using a USB stick, or
+  - download the ISO in Tails using the Unsafe Browser as follows:
+
+    - Start Tails with an administration password set in the Welcome Screen.
+    - Open the Unsafe Browser: **Applications > Internet > Unsafe Browser** and find and download the ISO.
+    - Leave the Unsafe Browser running, and open a terminal via **Applications > System Tools > Terminal**.
+    - Copy the ISO to the desktop with the command:
+
+      .. code-block:: sh
+
+        sudo cp /var/lib/unsafe-browser/chroot/home/clearnet/Downloads/<fileName.iso> ~amnesia/Desktop
+
+    - Fix the ISO file's ownership with the command:
+
+      .. code-block:: sh
+
+        sudo chown amnesia:amnesia ~/Desktop/<fileName.iso>
+
 - Verify the checksum of the downloaded ISO file using the following command, comparing it against the checksum in the file listing above:
 
   .. code-block:: sh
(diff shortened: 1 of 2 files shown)


eloquence:
(Rebased prior to review.)

docs/admin/install.rst
```diff
@@ -36,6 +36,9 @@ In order to install SecureDrop Workstation and configure it to use an existing S
 
 - The SecureDrop instance's Admin Workstation and Secure Viewing Station (SVS) USBs, and the full GPG fingerprint of the submission key.
 - A working computer (Linux is recommended and assumed in this guide) to use for verification and creation of the Qubes installation medium.
+
+  .. note:: A Tails USB can be used to perform the tasks below, but due to the size of the Qubes installation ISO, it may make sense to download it on another computer rather than via Tor, and then to use a uSB stick to transfer it to Tails for verification and creation of the installation medium.
```
eloquence, line 40: typo: usb->USB

docs/admin/hardware.rst
```diff
@@ -37,6 +37,28 @@ The instructions below assume the use of a Linux-based computer for the creation
...
+  - download the ISO in Tails using the Unsafe Browser as follows:
+
+    - Start Tails with an administration password set in the Welcome Screen.
+    - Open the Unsafe Browser: **Applications > Internet > Unsafe Browser** and find and download the ISO.
+    - Leave the Unsafe Browser running, and open a terminal via **Applications > System Tools > Terminal**.
+    - Copy the ISO to the desktop with the command:
+
+      .. code-block:: sh
+
+        sudo cp /var/lib/unsafe-browser/chroot/home/clearnet/Downloads/<fileName.iso> ~amnesia/Desktop
```
eloquence, line 54: I would shorten to `~/Desktop` (it's expanded pre-sudo)

docs/admin/hardware.rst
```diff
@@ -37,6 +37,28 @@ The instructions below assume the use of a Linux-based computer for the creation
...
 - Expand the **BIOS Update** listing and download the **BIOS Update (Bootable CD)** file.
+
+.. note::
+  A Tails USB can be used for the verification and conversion process described below, but the Lenovo support site blocks requests over Tor, preventing the ISO download. To work around this, either:
+
+  - download the BIOS ISO on a different computer and transfer it to Tails using a USB stick, or
+  - download the ISO in Tails using the Unsafe Browser as follows:
+
+    - Start Tails with an administration password set in the Welcome Screen.
+    - Open the Unsafe Browser: **Applications > Internet > Unsafe Browser** and find and download the ISO.
```
eloquence, line 48: I would add:

- Note down the name of the downloaded file.

(Because they'll otherwise have to `sudo ls` the directory contents later.)

docs/admin/install.rst
zenmonkeykstop: :heavy_check_mark:

docs/admin/hardware.rst
zenmonkeykstop: Imma leave it as-is, just to keep things explicit.

docs/admin/hardware.rst
zenmonkeykstop: :heavy_check_mark:

docs/admin/hardware.rst
eloquence: I'm fine either way, but note that you have it in the shorter form immediately below, which could be a bit confusing.

zenmonkeykstop: where is the collar-pull emoji when you need it! fixing.

PR #54, opened 2020-07-22
Update default branch name

eloquence:
Update README consistent with branch name change.


conorsch:
Still updating the version-controlled branch settings, self-assigning for review so they get merged together.

PR #58, opened 2020-09-04
add instructions around sys-whonix tor connection

creviera:
# Description

Add an instruction for selecting the "Connect" option to connect directly to the Tor network when sys-whonix prompts for this. 

A screenshot of the qubes dialog would be ideal, so if other agree, I can modify the sys-whonix qubes settings in order to get it to prompt me again and then take a screenshot.


eloquence:
LGTM.

conorsch:
Looks like this dialog popped up because on a fresh install, `whonix-gw-15` was updated for the first time, and `sys-whonix` was run for the first time. Certainly the docs should reference it. Let's add a screenshot next time someone's installing fresh.

conorsch (approved):

PR #59, opened 2020-09-13
fix typo

deeplow:


rocodes (approved):
Thanks for your keen eye, @deeplow! Appreciated as always.

PR #62, opened 2020-10-20
Provide documentation for Lenovo T490 support

### LINKED ISSUE #37: Test T490 and update hardware recommendations
The ThinkPad T480 is [no longer available](https://www.lenovo.com/us/en/laptops/thinkpad/thinkpad-t-series/ThinkPad-T480/p/22TP2TT4800) from Lenovo, which makes it unsuitable as an official hardware recommendation. The next product in the series is, unsurprisingly, the T490. We should do a full prod setup on that hardware (including BIOS changes) to ensure that it works, and document any needed workarounds.

emkll:
Closes #37 , also adds link to design document in FAQ


docs/admin/hardware.rst
```diff
@@ -22,18 +22,65 @@ In order to print submissions, a supported non-networked printer is required. Su
...
+-----------------------
+
+Lenovo ThinkPad T490 (with 8th generation Intel Core processor)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The Thinkpad T490 **with an 8th generation Intel Core processor** is a recommended option for the SecureDrop Workstation. If you plan to use it, you should follow the instructions below to ensure that the BIOS is up to date and adequately configured before proceeding with the installation.
+
+
+.. caution::
+
+  There versions of the T490 with 10th generation Intel Core processors, and are at present **untested and unsupported**.
```
eloquence, line 35: Typo/sentence construction issue ("there versions"):

> There versions of the T490 with 10th generation Intel Core processors, and are at present **untested and unsupported**.

How about:

> The versions of the T490 with 10th generation Intel Core processors are at present **untested and unsupported**.

eloquence:
(This, particularly the networking instructions, will require testing by a team member with an 8th gen T490, which other than Mickael I believe currently is only @conorsch.)

conorsch:
Happy to test on the T490 I've got. As long as there's no rush on merging these docs, I'll likely do that next sprint. Can reprioritize if we want confirmation before sharing more widely.

conorsch:
Discussed in standup today, I'll take a pass at installing on T490 to verify.

docs/admin/hardware.rst
```diff
@@ -22,18 +22,65 @@ In order to print submissions, a supported non-networked printer is required. Su
...
+
+
+For both device IDs (e.g. ``dom0:00_1f.6`` and ``dom0:00_14.3``), you will need to detach and re-attach the device to ``sys-net``, then restart ``sys-net`` as follows:
+
+.. code-block:: sh
+
+  qvm-pci detach sys-net dom0:00_14.3
+  qvm-pci detach sys-net dom0:00_1f.6
+  qvm-pci attach sys-net --persistent --option-no-strict-reset=True dom0:00_14.3
+  qvm-pci attach sys-net --persistent --option-no-strict-reset=True dom0:00_1f.6
```
conorsch, line 64: Based on testing just now, should be `s/option-/option=/`

emkll: good catch, fixed. Based on the message board post in the commit message i believe it should be `--option no-strict-reset=True`, fixed !

conorsch:
Followed these instructions on a fresh Qubes 4.0.3 install on a T490. They worked great! Both wifi & ethernet are working, and remain working after reboots. I'm going to make some language and formatting cleanups, as well as add model number if what I have locally is differently than what's documented in here already.

docs/admin/hardware.rst
conorsch: I agree! Added this suggestion as part of review.

conorsch (approved):
Worked like a charm. Thanks for preparing, @emkll!

PR #63, opened 2020-11-05
Recommend reboot after first install

conorsch:
Ensures that all AppVMs and TemplateVMs have been fully rebooted,
prior to first run. Related to QA findings in [0].

[0] https://github.com/freedomofpress/securedrop-workstation/issues/630


zenmonkeykstop (approved):
Looks good based on visual review.

PR #64, opened 2020-11-06
added temporary Salt pinning instructions for fedora-31

zenmonkeykstop:
Adds a section in the `fedora-31` installation instructions, cribbed from the workaround in https://github.com/freedomofpress/securedrop-workstation/issues/485#issuecomment-595244676 , to pin the salt and salt-ssh packages in `fedora-31` at their initial versions.

This works around issue https://github.com/QubesOS/qubes-issues/issues/6188, but should be removed once a fix exists.

### DIFF
--- a/docs/admin/upgrading_to_fedora_31.rst
+++ b/docs/admin/upgrading_to_fedora_31.rst
@@ -40,6 +40,26 @@ bar.
 When the download has concluded, you will be prompted to install the package.
 Type ``y`` to proceed with the installation.
 
+TEMPORARY: Exclude Salt packages from updates
+---------------------------------------------
+
+.. important:: Due to an `upstream issue <https://github.com/QubesOS/qubes-issues/issues/6188>`_,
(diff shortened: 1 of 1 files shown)


conorsch:
On local review, the call-out box isn't actually rendering:

f31-workaround-not-showing

So we'll need to reformat it a bit. Worth considering whether to add an equally temporary mention in FAQ, but I assume the addition here will give us all we need to link out until the problem is resolved upstream.

conorsch:
My mistake, it's showing up just fine, I was looking in the wrong spot:

f31-update-definitely-showing

Have minor edits to language, will append & approve, then toss back to you for a second look @zenmonkeykstop

conorsch (approved):
Looks good. We'll monitor changes upstream and revise accordingly. 

@zenmonkeykstop Please take a look at the small tweaks I added, and merge if you agree.

zenmonkeykstop:
LGTM, merging!

PR #65, opened 2020-11-10
Documents use of test repos for qubes-mgmt-salt package

zenmonkeykstop:
Includes the updated steps to pull in the test version of `qubes-mgmt-salt-vm-connector`,  as a more reasonable workaround for https://github.com/QubesOS/qubes-issues/issues/6188.

## Testing:
- [ ] CI is passing
- [ ] docs are clear and correct
- [ ] steps work on a Qubes system that does not already have downgraded and pinned versions of the `salt,salt-ssh` packages.


eloquence:
LGTM. I've followed those same steps on my machine yesterday (after reverting the downgrade), and they appear to be working as expected.

eloquence (approved):

PR #72, opened 2021-03-06
Add pre-install task to rotate legacy passphrases

### LINKED ISSUE #71: Document passphrase requirements for SecureDrop accounts
The use of diceware passphrases in SecureDrop was enforced via https://github.com/freedomofpress/securedrop/pull/1509, which was released with [SecureDrop 0.4.3](https://securedrop.org/news/securedrop-043-released/) in September 2017. That PR also increased the minimum character length for passphrases to 14 characters (from 12).

The SecureDrop Client enforces a 14 character minimum as well. That means it is not currently possible to log in with legacy accounts that have shorter passwords or passphrases. 

Unless we change that restriction in the SecureDrop Client, we should at minimum document that older passwords/passphrases must be rotated.

eloquence:
Ready for review.

Resolves #71 

This is a very conservative approach that may involve rotation on some accounts that don't require it, but I think erring on the side of a simple procedure makes sense for the documentation; if folks want to discuss in more detail vis-a-vis an organization's specific procedures, that can be done in support.

### DIFF
--- a/docs/admin/install.rst
+++ b/docs/admin/install.rst
@@ -54,7 +54,7 @@ Pre-install tasks
 
 Verify the SecureDrop server configuration
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


conorsch (approved):
:+1:

conorsch (approved):
Re-approving after rebase to satisfy staleness check.

PR #73, opened 2021-03-08
Revise docs for Qubes 4.0.4

eloquence:
- Updated version number
- Installation of fedora-32 no longer required

## Review notes

Visual review / comparison with release notes at https://www.qubes-os.org/news/2021/03/04/qubes-4-0-4/ should be sufficient for this PR


conorsch (approved):
Looks great. We'll have to revive these docs from the history as early as May 2021, when it's time to upgrade Fedora, but we'll cross that bridge when we come to it.

PR #77, opened 2021-03-26
Unpin pip and pip-tools and bump requirements

eloquence:
It's not clear that there's a benefit in specifying a version of `pip` and `pip-tools` for these docs build environments in the `requirements.in` file, and we don't do so over in `securedrop-docs`: 
https://github.com/freedomofpress/securedrop-docs/blob/main/requirements/requirements.in

Instead, this PR bumps all requirements to latest. Once merged, #75 and #76 can be closed.

The syntax used to bump the requirements was the following:
`pip-compile --allow-unsafe --generate-hashes --output-file=requirements.txt requirements.in`

## Test plan

- Create a virtualenv with a [supported](https://endoflife.date/python) version of Python and install the requirements.
- [ ] Observe that `Makefile` targets are still working as expected.


zenmonkeykstop (approved):
Make targets work fine with a fresh python 3.7 virtualenv. Also tested the RTD build by temporarily activating (and hiding) this branch. Builds fine. Good to go.

Issue #46, opened 2020-05-31
Document backup/restore procedures


eloquence:
Clear backup/restore procedures are going to be a must-have requirement for wider production usage of SecureDrop Workstation.

eloquence:
As a first step, @rocodes has offered to spend about a day on exploring the current options. We're interesting in questions like:

- Can a working SecureDrop Workstation be reliably backed up and fully restored using Qubes' own backup tools?
- If not, are there alternatives we should consider for a full system backup?
- If a full system backup/restore presents significant challenge, can we facilitate restoring data/configuration upon a reinstall?

This will guide our documentation efforts.

eloquence:
@rocodes Can you post a summary of your findings so far to this issue?

eloquence:
Cross-referencing progress update here: https://github.com/freedomofpress/securedrop-workstation/issues/120#issuecomment-654877778

rocodes:
Circling back to this: 
- As per https://github.com/freedomofpress/securedrop-workstation/issues/120, a combination of the existing Qubes backup tooling, plus some additional tooling that backs up the required components from `dom0` (if we decide we want to do this), would allow for a good backup-and-restore strategy, but see below note on security.
- The sdw backup/restore story should allow afford backing up custom qubes, most likely using the qubes gui backup tool, for users that customize the Workstation 
- Whatever backup tooling we create should be reviewed carefully (see for example, [paranoid restore mode](https://www.qubes-os.org/news/2017/04/26/qubes-compromise-recovery/#introducing-the-paranoid-backup-restore-mode), which doesn't restore anything to dom0). SecureDrop Workstation relies on settings (such as rpc policies and/or tags) that go beyond the scope of what Qubes would normally back up or restore to dom0.  
- I think it's reasonable to break this into 2 goals: a quick/simple first pass at a backup and restore story (likely a partially manual process) and some docs, and then a followup at some point later on with a more developed tool (e.g. with gui integration) if we're happy with the workflow. 

### Scope of backup questions 
- Do we back up:
   -  `sd-app`, `sd,dev`, `sd-gpg`, `sd-log`, `dom0`?  
   - appVms, all templates, `dom0`?
   - appVMs, templates, `dom0`, `fedora32 template`? (least sensible imo)
- Do we back up and restore `dom0`:
    - home directory, per current Qubes policy)? 
[...]

PR #81, opened 2021-05-21
Add backup and restore instructions

rocodes:
**Status** WIP/Ready for initial review
**Description** Document backup and restore process 
**Resolves** #46 
**Related** https://github.com/freedomofpress/securedrop-workstation/issues/120

The approach taken in this PR is to suggest backing up the vault vm and the configuration files from `dom0`, and nothing else. (See comms in linked issues above). 

I hope to improve on this approach eventually, because it's still a little unwieldy. Open questions include:
- Re-downloading and verifying the rpm and then shuttling it into dom0 seems like a real pain, but relying on a (potentially _very_ old) rpm + the updater also doesn't seem great. Currently, to be conservative, I recommend users re-download/verify/shuttle the RPM if it's not the current version. Open to feedback. 
- Currently, if users have configured custom RPC policies (such as tags that grant extended clipboard access), they have to manually re-add those after restoring the backup. Rationale: overwriting `/etc/qubes-rpc` in dom0 is risky, shouldn't be undertaken lightly, and has potential for breakage or bad consequences (for example, if VM names change and a hardcoded rpc policy is in place, or if a VM is granted access in an RPC policy but isn't restored in the backup, then if a future VM is created with that name, it could have unexpected permissions grandfathered in...yknow, stuff to be cautious/paranoid about). Again, tried to be conservative here, open to feedback.

I think it will make sense to have a much more streamlined backup/restore, but this is a start. (Future ideas: maybe a "bundle backup configs" button in the GUI,  maybe an easier way to grab/verify the RPM, and maybe a different consensus on whether or not we backup and restore more of the system, such as templates, so as to avoid long wall times/manual intervention during restore?)

### DIFF
--- a/docs/admin/backup.rst
+++ b/docs/admin/backup.rst
@@ -8,54 +8,53 @@ that allows for backup and restoration of user-specified VMs.
 
 To perform backups, you will need:
 
- - an encrypted storage medium
- - a secure place to store backup credentials    
+ - a LUKS-encrypted external hard drive, with at least 50GB space
+ - a secure place to store backup credentials (such as a password manager
+   on your primary laptop)    
 
 Backup 
 ------
 
-Prepare files in ``dom0``
-~~~~~~~~~~~~~~~~~~~~~~~~~
+Preserve files from ``dom0``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Files outside the ``dom0`` home folder aren't normally backed up by the 
-Qubes backup tool. Preserve key configuration files by coping them to the 
-home directory. 
+Preserve key configuration files by coping them into the 
+``vault`` VM.
 
 In a ``dom0`` Terminal via **Q ▸ Terminal Emulator**:
 
   .. code-block:: sh
 
-    mkdir ~/backup/
-    cp -r /usr/share/securedrop-workstation-dom0-config/ ~/backup/
+    cd /usr/share/securedrop-workstation-dom0-config
+    qvm-copy-to-vm vault {config.json,sd-journalist.sec}
 
-If you have customized :doc:`clipboard access <managing_clipboard>` 
-to SecureDrop Workstation, or have made any other customizations 
-involving RPC policies, back up related configuration files:   
+Open a ``vault`` Terminal and verify that the files were copied successfully:
 
   .. code-block:: sh
+  
+    head -n1 ~/QubesIncoming/dom0/sd-journalist.sec # line contains "BEGIN PRIVATE KEY BLOCK"
+    grep -q descriptor ~/QubesIncoming/dom0/config.json && echo OK # line is "OK"
 
-   mkdir -p ~/backup/etc/qubes
-   mkdir ~/backup/etc/qubes-rpc
-   cp -a /etc/qubes/* ~/backup/etc/qubes/
-   cp -a /etc/qubes-rpc/* ~/backup/etc/qubes-rpc
+.. note::
+  If you have made advanced customizations to your Qubes Workstation, 
+  you may need to back up additional components of ``dom0``. Refer to
+  the `Qubes documentation <https://www.qubes-os.org/doc/backup-restore/>`_
+  or contact Support.
 
 Back up SecureDrop Workstation
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Ensure your storage medium is plugged in, attached to a VM, and mounted.
-(LUKS-encrypted storage media may be attached to ``sd-usb``, or you may use a custom 
-VM of your choosing, if you are relying on another form of encryption such as 
-Veracrypt).
+Ensure your storage medium is plugged in, attached to ``sd-devices``,
+and unlocked.
 
 Navigate to **Q ▸ System Tools ▸ Backup Qubes**, and move all VMs from 
 "Selected" to "Available" by pressing the ``<<`` button. 
 
 To target a VM for backup, highlight it and move it into the "Selected" 
-column by pressing the ``>`` button. Select the following VMs:
+column by pressing the ``>`` button. Select:
 
- - ``dom0``
- - ``vault``
- - any customized VMs that you wish to preserve.
+- the ``vault`` VM
+- any customized VMs that you may wish to preserve.
 
 You do not need to back up the ``sd-`` VMs.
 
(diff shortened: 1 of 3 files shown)


docs/admin/backup.rst
```diff
@@ -0,0 +1,177 @@
...
+column by pressing the ``>`` button. Select the following VMs:
+
+ - ``dom0``
+ - ``vault``
+ - any customized VMs that you wish to preserve.
+
+You do not need to back up the ``sd-`` VMs.
+
+Click "Next", and in "Backup destination," specify the VM and directory 
+corresponding to your storage medium's current mount point.
```
conorsch, line 63: I've got light concerns about how little detail we're providing on the mechanics of creating a LUKS-encrypted volume, mounting it, and unlocking it. My personal backup procedures are roughly:

1. create `backup` appvm with no network
2. connect external hard drive , use sys-usb to connect it to `backup`
3. run cli `crytsetup luksOpen` and `mount` commands (bleh)
4. then switch back to dom0 and use the backup utility

I don't think we need to go into that much detail in our docs, it's much better to link out to upstream docs, as you've done above. It'd be an interesting exercise to ask someone without a ton of Qubes familiarity to run through these backup docs. Maybe we can try that at a later stage of review. =)

conorsch: I would expect a LUKS-encrypted drive to be automatically attached to `sd-devices` on fully provisioned Workstation. If that's the case, fine to use that to perform the backup!

docs/admin/backup.rst
```diff
@@ -0,0 +1,177 @@
...
+Reinstall QubesOS
+~~~~~~~~~~~~~~~~~
+
+To restore SecureDrop Workstation, follow our 
+:doc:`pre-install tasks <install>` to provision a QubesOS system complete with
+updated base templates.
+
+.. note::
+ If you backed up a number of custom VMs, during the installer, you may opt
+ to un-select "create default application qubes (personal, work, untrusted, vault)".
```
conorsch, line 92: To reduce the number of "ifs", maybe we can just say explicitly here _don't_ install those default VMs. We don't care about personal, work, or untrusted—and vault will be coming from the backup.

docs/admin/backup.rst
```diff
@@ -0,0 +1,177 @@
...
+   mkdir -p ~/backup/etc/qubes
+   mkdir ~/backup/etc/qubes-rpc
+   cp -a /etc/qubes/* ~/backup/etc/qubes/
+   cp -a /etc/qubes-rpc/* ~/backup/etc/qubes-rpc
+
+Back up SecureDrop Workstation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Ensure your storage medium is plugged in, attached to a VM, and mounted.
+(LUKS-encrypted storage media may be attached to ``sd-usb``, or you may use a custom
```
conorsch, line 46: `s/sd-usb/sd-devices/`

docs/admin/backup.rst
conorsch: Ah, right: above it was referred to as `sd-usb`, and I'd planned to re-target that as `sys-usb`. Looks like `sd-devices` is what we want, and it's already addressed in a comment above.

docs/admin/backup.rst
```diff
@@ -0,0 +1,177 @@
...
+The contents of the previous ``dom0`` will be backed up to a folder in the 
+new ``dom0`` home directory called ``home-restore-<timestamp>/dom0-home/user``.
+In this guide we will refer to that location as ``$DOM0_BACKUP``.
+
+Open a terminal via **Q ▸ Terminal Emulator**: 
+
+  .. code-block:: sh
+
+   cd $DOM0_BACKUP/backup
+   cp -r securedrop-workstation-dom0-config/ /usr/share/
```
conorsch, line 129: :rotating_light: See note in backup section about the potentially harmful recursive copy here—we only want two (2) files. It's most important to specify them explicitly by fullpath in the backup section, but I'd still do the same in the restore section, to guard against surprises.

docs/admin/backup.rst
```diff
@@ -0,0 +1,177 @@
...
+
+Follow the :doc:`installation instructions <install>` to download and verify
+the SecureDrop Workstation rpm to a network-attached VM and copy it to ``dom0``.
+
+.. note::
+ If your installation is recent, your backup archive may contain the current
+ version of the rpm, in which case you can skip the above steps and move
+ straight to installing the rpm. 
+ Use ``rpm -qi securedrop-workstation.rpm`` to compare the version with the
+ latest version at ``https://yum.securedrop.org/workstation/dom0/f25/``
```
conorsch, line 152: In the interest of reducing `if`s—which I find cause me to stop reading and think hard about whether the condition applies to my use case—let's just recommend installing from scratch again. It feels like an edge case for the back up and restore to happen on the same version of `securedrop-workstation-dom0-config`. Even in the case of a minor version bump, we want to make sure the new system uses the latest and greatest.

docs/admin/backup.rst
```diff
@@ -0,0 +1,177 @@
...
+Click "Next", and in "Backup destination," specify the VM and directory 
+corresponding to your storage medium's current mount point.
+
+Set a strong, unique backup passphrase (7-word diceware), and ensure this 
+passphrase is stored securely outside SecureDrop Workstation. 
+
+.. note:: 
+ This passphrase protects sensitive 
+ components of your SecureDrop instance, including the *Submission Private Key*.
+ Ensure that it is a very strong password and is stored securely.
```
conorsch, line 71: Side note: we should be installing `diceware` in the SDW templates and instructing admins to use it.

docs/admin/backup.rst
```diff
@@ -0,0 +1,177 @@
...
+
+.. include:: ../includes/top-warning.rst
+
+QubesOS has a `backup utility <https://www.qubes-os.org/doc/backup-restore/>`_ 
+that allows for backup and restoration of user-specified VMs. 
+
+To perform backups, you will need:
+
+ - an encrypted storage medium
+ - a secure place to store backup credentials
```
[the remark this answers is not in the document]
conorsch: Similiarly, consider 

> a secure place to store backup credentials (such as a password manager on your primary laptop)

conorsch (commented):
Very nice! Left extremely verbose comments throughout. Basically my feedback boils down to 

1. let's ignore most of what's in dom0
2. the fresh install pre-restore should be as pristine as possible, including latest version

Note that I haven't actually run through these instructions to verify the steps functionally. After you've had a chance to review the comments, please ping me again for a deeper review!

docs/admin/backup.rst
```diff
@@ -0,0 +1,177 @@
...
+
+Navigate to **Q ▸ System Tools ▸ Backup Qubes**, and move all VMs from 
+"Selected" to "Available" by pressing the ``<<`` button. 
+
+To target a VM for backup, highlight it and move it into the "Selected" 
+column by pressing the ``>`` button. Select the following VMs:
+
+ - ``dom0``
+ - ``vault``
+ - any customized VMs that you wish to preserve.
```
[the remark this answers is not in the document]
rocodes: I agree with you that I don't personally restore dom0, but here are the things that I wonder about w.r.t dom0:
- ease of workflow for a user that has configured copy-paste between VMs or other custom RPC policies (very common eg with passwords and the vault vm)
- _if_ just doing a simple hardware migration for some reason, and a user already has the latest RPM in dom0, it seems rude to make them go through the whole 'download, verify, shuttle to dom0 again' procedure.

I'm in favour of rethinking where those files are stored. I think we can safely expect that they will live on in perpetuity wherever they are copied to, so we have to take that into account, esp with `sd-journalist.sec`.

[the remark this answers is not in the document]
conorsch: >  user that has configured copy-paste between VMs or other custom RPC policies, possibly with our assistance (very common eg with passwords and the vault vm)

Backing up dom0 won't help in that case. In order to support retaining clipboard access settings, we'd have to either 1) backup the `sd-*` VMs (which would include tag info for each VM); or 2) write a script to dump the tags from the SDW VMs, and then encourage the admin to re-apply the tags after restoring. Neither seem worthwhile to my eye, although at least mentioning the clipboard access tags in the backup/restore docs certainly seems reasonable.

> if just doing a simple hardware migration for some reason, and a user already has the latest RPM in dom0

That strikes me as a lot of conditionals. In order to reduce headaches for admins, I think our efforts would be best spent in trying to avoid the shuttle-to-dom0 workflow entirely, e.g. via `--enablerepo=qubes-templates-community`. That'll take a lot of work in coordination with Qubes upstream, so definitely a long-term goal. I defer to your judgment on what's most helpful to admins at present!

docs/admin/backup.rst
```diff
@@ -0,0 +1,177 @@
...
+In a ``dom0`` Terminal via **Q ▸ Terminal Emulator**:
+
+  .. code-block:: sh
+
+    mkdir ~/backup/
+    cp -r /usr/share/securedrop-workstation-dom0-config/ ~/backup/
+
+If you have customized :doc:`clipboard access <managing_clipboard>` 
+to SecureDrop Workstation, or have made any other customizations 
+involving RPC policies, back up related configuration files:
```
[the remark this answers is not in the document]
rocodes: In my experience, tags are restored when a qube is restored. (I remembered saying this [a while back](https://github.com/freedomofpress/securedrop-workstation/issues/120#issuecomment-656918934), but I was doubting myself, so I just tried tagging and restoring a test qube and indeed the custom tags were still present).  

Backing up the RPC policy directories (with the idea that admins could audit them if need be, but not restore them directly) was some sort-of-futureproofing in case we are dealing with customized export workflows that involve customized qubes. But I can remove it and we can cross that bridge if/when we come to it.

docs/admin/backup.rst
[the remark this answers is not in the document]
rocodes: Resolved in 05b8dfd, where I no longer suggest backing up `dom0`.

docs/admin/backup.rst
rocodes: thanks for this flag-- fixed in 05b8dfd

docs/admin/backup.rst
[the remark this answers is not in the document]
rocodes: I've removed all of this in 05b8dfd and left a note that advanced users may need to back up additional `dom0` components. I think for now, that's a reasonable tradeoff information-wise. If we have orgs using a lot of customized qubes and/or modifying RPC policies to accommodate their own workflows down the line, we can address this in more depth (or maybe we will develop a system that avoids the need for them doing this).

docs/admin/backup.rst
```diff
@@ -0,0 +1,177 @@
...
+If you have customized :doc:`clipboard access <managing_clipboard>` 
+to SecureDrop Workstation, or have made any other customizations 
+involving RPC policies, back up related configuration files:   
+
+  .. code-block:: sh
+
+   mkdir -p ~/backup/etc/qubes
+   mkdir ~/backup/etc/qubes-rpc
+   cp -a /etc/qubes/* ~/backup/etc/qubes/
+   cp -a /etc/qubes-rpc/* ~/backup/etc/qubes-rpc
```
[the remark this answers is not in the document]
rocodes: removed in 05b8dfd

docs/admin/backup.rst
rocodes: oh, I remember why I had the "if" and this workflow here -- if they clean install and then need to re-download the RPM, without the default application qubes, they don't actually have a network-connected VM with which to grab the RPM, so our instructions are technically incomplete.

rocodes: (fixed in c462517)

rocodes:
@conorsch, thanks for your review! I've addressed your comments and now the backup/restore workflow involves:
- backup: copy `sd-journalist.sec` and `config.json` to the vault vm; back up vault vm

- restore: reinstall qubes, restore vault, re-create work VM, re-download RPM, copy in and verify config files from vault vm, run `apply`.

docs/admin/backup.rst
```diff
@@ -0,0 +1,156 @@
...
+
+Optionally, inspect each file before proceeding. The first 
+file should be an ASCII-armored GPG private key file, and the second is a 
+one-line file with the format ``ONIONADDRESS:descriptor:x25519:AUTHTOKEN``.
+
+Copy both files into place:
+
+  .. code-block:: sh
+
+    sudo cp tmp{sd-journalist.sec,config.json} /usr/share/securedrop-workstation-dom0-config/
```
conorsch, line 137: Missing leading slash on `tmp`, should be `/tmp/`.

conorsch:
Thanks for the copious edits, @rocodes! Looks great to my eye. Before merge, I'd like to see @creviera take a pass through it so it's got her formal approval. No further concerns from me.

creviera:
on my radar: I can review today or tomorrow

docs/admin/backup.rst
creviera: confirmed this works as expected and documented

docs/admin/backup.rst
creviera: created followup issue

creviera (dismissed):
lgtm, thanks ro!

creviera (approved):
looks great! :rocket:

PR #82, opened 2021-05-21
Update signing key

eloquence:
Part of the 2021 signing key rotation epic (https://github.com/freedomofpress/securedrop/issues/5923).

This is blocked until we start signing releases with the new key, so draft for now.

### DIFF
--- a/docs/admin/install.rst
+++ b/docs/admin/install.rst
@@ -119,7 +119,7 @@ Download and verify Qubes OS
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 On the working computer, download the latest stable Qubes OS ISO (``4.0.4`` at time of writing) from `https://www.qubes-os.org/downloads/ <https://www.qubes-os.org/downloads/>`_. The ISO is 4.5GB approximately, and may take some time to download based on the speed of your Internet connection.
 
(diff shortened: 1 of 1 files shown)


rocodes:
Since key rotation changes are in, if this is ready for review I can do so!

eloquence:
The current RPM that's live on https://yum.securedrop.org/workstation/dom0/f25/ is still signed with the old key (ref: https://github.com/freedomofpress/securedrop-workstation-prod-rpm-packages-lfs/pull/19). `rpm` still successfully validates it in spite of the key expiry, but we can't update this part of the docs until we release a new RPM signed with the new key.

conorsch:
Once https://github.com/freedomofpress/securedrop-workstation-prod-rpm-packages-lfs/pull/21 is merged, we can review these docs and get them in.

conorsch:
https://github.com/freedomofpress/securedrop-workstation-prod-rpm-packages-lfs/pull/21 is in! So I'm marking this ready for review. Soon I'll run through a clean prod install, following along with the docs, to verify everything's as it should be.

eloquence:
(CI is a linkcheck failure, which we should probably break out into nightly as we do in the main docs repo. But will append a commit to fix.)

conorsch (approved):
This looks great. I performed a clean prod install to verify, and while I encountered problems (https://github.com/freedomofpress/securedrop-workstation/issues/721), those are tracked separately and shouldn't block merge here.

PR #88, opened 2021-10-22
Update all requirements

eloquence:
Addresses vulnerabilities noted in #83 and #87

To test, install in virtualenv and run `make docs`.


.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2
 jobs:
   lint:
     docker:
-      - image: python:3.7
+      - image: cimg/python:3.10.0
```
eloquence, line 6: Granted, the freshest version we use anywhere, but I figure these easy-to-fix repos with low technical complexity are a good place to take new images for a spin.

PR #91, opened 2021-11-15
Update fedora-33 -> fedora-34 throughout

conorsch:
Towards https://github.com/freedomofpress/securedrop-workstation/issues/735

Makes the usual string substitutions. Also renames the fedora RST file so the URL is stable over time.

### Testing
1. Visual review
2. `make docs` locally and ensure the renamed file is still reachable via clickable URLs in all the right places


creviera (approved):
visual review looks good and renamed page is accessible

PR #96, opened 2021-12-22
Add k8s container

maeve-fpf:
This PR is the same as https://github.com/freedomofpress/securedrop-docs/pull/295; please see there for an explanation. The changes are somewhat smaller as there are fewer LaTeX considerations here, and I squashed some parts into fewer commits.

The staging deployment for this one is: https://staging-workstation-docs.securedrop.org/

Like `securedrop-docs`, CI is currently set up to trigger on a test branch of this repo, so nothing will happen immedately when this is merged into `main`.

To verify that the content is different between stable and latest, you can check `admin/install.html` - it will refer to Fedora 33 or 34.


maeve-fpf:
I've added a commit here, branched off of `stable`, as a candidate for that tag. It contains a subset of these changes, squashed into one commit, so that a build with that tag checked out will generate a version switcher and PDF: https://github.com/freedomofpress/securedrop-workstation-docs/compare/stable...stable-plus-k8s-container

eloquence (changes requested):
The build currently errors out for me due to LaTeX issues:

```
! Package inputenc Error: Unicode character ▸ (U+25B8)
(inputenc)                not set up for use with LaTeX.
```

I see that the character is declared, haven't investigated much further yet.

eloquence (approved):
Per chat with @maeve-fpf , re-tested from `staging-ci-test` branch which is identical except for pointing to branches configured for successful testing. Works as expected, `stable` tag is honored.

PR #97, opened 2022-01-12
documents ThinkPad T14 compatibility

cfm:
Closes freedomofpress/securedrop-workstation#749 by documenting:

1. hardware compatibility, BIOS updates, and Qubes settings for the ThinkPad T14; and
2. a few extra steps necessary as of January 2022 now that both Fedora 32 and Whonix 15 are obsolete on arrival in new Qubes 4.0.4 installations.

### DIFF
--- a/docs/admin/hardware.rst
+++ b/docs/admin/hardware.rst
@@ -17,12 +17,21 @@ More information on hardware compatibility can be found on the `Qubes OS System 
 
 In order to print submissions, a supported non-networked printer is required. We have tested and recommend the HP LaserJet Pro M404n. More printer options will be added in future releases.
 
-Lenovo T series Laptops
+.. _thinkpad_t_series:
+
+Lenovo T series laptops
 -----------------------
 
-Lenovo ThinkPad T490 (with 8th generation Intel Core processor)
+Lenovo ThinkPad T14 (with 10th-generation Intel Core processor)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-The Thinkpad T490 **with an 8th generation Intel Core processor** is a recommended option for the SecureDrop Workstation. If you plan to use it, you should follow the instructions below to ensure that the BIOS is up to date and adequately configured before proceeding with the installation.
+The ThinkPad T14 **with a 10th-generation Intel Core processor** is a recommended option for the SecureDrop Workstation. If you plan to use it:
+
+- If your laptop has come with Ubuntu preinstalled, run its **Software Updater** to update the BIOS automatically (via ``fwupd``).
+- Otherwise, follow the instructions below to ensure that the BIOS is up to date.
+
+Lenovo ThinkPad T490 (with 8th-generation Intel Core processor)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The Thinkpad T490 **with an 8th-generation Intel Core processor** is a recommended option for the SecureDrop Workstation. If you plan to use it, you should follow the instructions below to ensure that the BIOS is up to date and adequately configured before proceeding with the installation.
 
 
 .. caution::
--- a/docs/admin/hardware.rst
+++ b/docs/admin/hardware.rst
@@ -26,7 +26,13 @@ Lenovo ThinkPad T14 (with 10th-generation Intel Core processor)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 The ThinkPad T14 **with a 10th-generation Intel Core processor** is a recommended option for the SecureDrop Workstation. If you plan to use it:
 
-- If your laptop has come with Ubuntu preinstalled, run its **Software Updater** to update the BIOS automatically (via ``fwupd``).
+- If your laptop has come with Ubuntu preinstalled, run its **Software Updater** twice as follows:
+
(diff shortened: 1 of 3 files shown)


docs/admin/hardware.rst
```diff
@@ -17,12 +17,25 @@ More information on hardware compatibility can be found on the `Qubes OS System
...
+
+Lenovo T series laptops
 -----------------------
 
-Lenovo ThinkPad T490 (with 8th generation Intel Core processor)
+Lenovo ThinkPad T14 (with 10th-generation Intel Core processor)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The ThinkPad T14 **with a 10th-generation Intel Core processor** is a recommended option for the SecureDrop Workstation. If you plan to use it:
+
+- If your laptop has come with Ubuntu preinstalled, run its **Software Updater** to update the BIOS automatically (via ``fwupd``).
```
legoktm, line 29: When I booted into the Ubuntu that came with my T14, there was an outstanding update for fwupd that I applied first before running it. I didn't check what the already-installed version was (and I already wiped the Ubuntu partition unfortunately), but based on https://launchpad.net/ubuntu/+source/fwupd (if you click on the package versions for focal, it'll show you the changelog entries since the previous version) it might have been a version vulnerable to CVE-2020-10759 (signature verification bypass). So if it's not too complicated, I think we should recommend running normal package updates first to get the latest `fwupd` version and then running `fwupd`.

docs/admin/hardware.rst
```diff
@@ -17,12 +17,25 @@ More information on hardware compatibility can be found on the `Qubes OS System
...
+- Otherwise, follow the instructions below to ensure that the BIOS is up to date.
+
+Network devices (Ethernet and Wi-Fi) will not immediately work out of the box and require a one-time manual configuration on install. After Qubes starts for the first time, when ``sys-net`` fails to start, follow the instructions below for the :ref:`thinkpad_t490`.
```
legoktm, line 32: I only had to follow this procedure for the ethernet controller (`dom0:00_1f.6`), but I also installed Qubes 4.1, which has a newer kernel. Just to make sure, did you need to do it for both the wifi and ethernet devices on Qubes 4.0?

docs/admin/hardware.rst
cfm: Good catch, @legoktm.  Let me know what you think of the more-detailed instructions proposed in d41a49ff9c434e034f3c4962a2081f76b2cb9d79.

docs/admin/hardware.rst
cfm: This was my experience too; thanks for corroborating.  Clarified in ad6710183afe135216388aed55726f8fd895bc08.

legoktm (dismissed):

legoktm:
LGTM, but I think you will want to rebase it before merging.

cfm:
Thanks, @legoktm!  This now just needs a maintainer's sign-off for merge.

legoktm (approved):

docs/admin/install.rst
```diff
@@ -305,10 +311,11 @@ With the key and configuration available in ``dom0``, you're ready to set up Sec
...
 
     securedrop-workstation-dom0-config-<versionNumber>-1.fc25.noarch.rpm:
-      Header V4 RSA/SHA256 Signature, key ID 7b22e6a3: OK
+      Header V4 RSA/SHA512 Signature, key ID 7b22e6a3: OK
+      Header SHA256 digest: OK
       Header SHA1 digest: OK
-      V4 RSA/SHA256 Signature, key ID 7b22e6a3: OK
-      MD5 digest: OK
+      Payload SHA256 digest: OK
+      V4 RSA/SHA512 Signature, key ID 7b22e6a3: OK
```
eloquence, line 318: I also see `MD5 digest: OK` in my output - if that's consistently part of the output I think we should include it here to match the full output.

docs/admin/hardware.rst
```diff
@@ -17,12 +17,31 @@ More information on hardware compatibility can be found on the `Qubes OS System
...
+The ThinkPad T14 **with a 10th-generation Intel Core processor** is a recommended option for the SecureDrop Workstation. If you plan to use it:
+
+- If your laptop has come with Ubuntu preinstalled, run its **Software Updater** twice as follows:
+
+  #. to install software updates, especially for the ``fwupd`` package; and then
+  #. to run ``fwupd`` to update the BIOS automatically.
+
+  If **Software Updater** offers to run ``fwupd`` during step (1), decline until step (2), to make sure ``fwupd`` itself has received its latest security updates.
+
+- Otherwise, follow the instructions below to ensure that the BIOS is up to date.
```
eloquence, line 36: Those instructions still specifically refer to T480/T490; have you had a chance to confirm that the process is essentially the same? If so perhaps we can generalize them a bit (can be done in follow-up IMO, but just noting).

eloquence (changes requested):
One nit and one question, otherwise LGTM.

docs/admin/install.rst
cfm: Good catch, @eloquence.  Fixed in 441cc36bccdd681e7382e5d53bafab0f512286e0.

docs/admin/hardware.rst
cfm: Good idea.  I'll tack on a commit proposing this generalization for your re-review on Monday.

cfm: Added in 26fb6dc.

eloquence (approved):
LGTM! In related news, it looks like 1) the supply of 10th gen T14 is dwindling and the ones currently available have a lower RAM ceiling (up to 16 GB), 2) the 11th gen models can no longer be customized with Ubuntu preinstalled. :/

PR #98, opened 2022-01-13
Remove apache header checks

conorsch:
The HTTP header checks were necessary in case an admin hadn't run a
playbook since those changes were added. As of 2021-04, all admins have
run a playbook, due to the Ubuntu LTS migration to 20.04. We can safely
assume the Apache config is sufficiently up to date.


eloquence (approved):
Thanks much, good catch!

PR #102, opened 2022-02-07
Add Qubes 4.1 (in-)compatibility note

eaon:
While we're working on a release that's compatible with R4.1, we want to make sure users keep installing Qubes OS 4.0.4. This also changes the link anchor to lead directly to the 4.0.4 release.


conorsch (approved):
Thanks, @eaon!

Issue #105, opened 2022-04-27
Docs update for 4.1: screenshots


eloquence:
The GUI has changed in some significant and some more subtle was on 4.1 (e.g., very significant changes to VM iconography). We should do a full screenshots refresh around the time of the grand migration.

cfm:
I've reviewed the remaining pre‒Qubes 4.1 screenshots.  My instinct is that we'd be best served by doing further updates to reflect the SecureDrop Client's current GUI after its next release.  Please feel free reopen this ticket, @eloquence or @nathandyer, if you feel these updates deserve attention sooner than that!

PR #113, opened 2022-07-07
update priority screenshots for Qubes 4.1

cfm:
Towards #105, updates priority screenshots—i.e., those that will be helpful for upcoming guided installations with pilot participants—for Qubes 4.1, per <https://github.com/freedomofpress/securedrop-workstation-docs/issues/105#issuecomment-1178076726>.

## Checklist

- [ ] In GitHub's visual diff, screenshots are correct and identical except for the intended differences between Qubes 4.0 and 4.1.


eloquence (approved):
Looks good! Merging without forwarding `stable` tag until the main 4.1 docs PR lands.

A minor thing we may want to standardize going forward is the username used in screenshots. I used `nbly` as a fake journalist name, but we should probably just standardize on `user` since that's what we recommend using in the install docs.

PR #114, opened 2022-07-11
Fix config check instructions in backup/restore guide

eloquence:
`descriptor` does not appear in config.json, `hostname` does.


rocodes (approved):
Thank you @eloquence! :)

PR #115, opened 2022-07-12
Update docs for Qubes 4.1.1

eloquence:
## Description

Qubes 4.1.1 ships with fedora-36, which for now means we can skip
installing an updated template in the installation phase. We will
need to reinstate this chapter after Fedora 36 hits EOL
(2023-05-16), assuming Qubes doesn't issue another point release
to support the next version.

## Status

Ready for review


eaon (approved):
Yay for speeding up installs at least until Fedora 37 rears its head ;)

PR #119, opened 2022-07-19
Clarifications to installation and restore docs

eloquence:
- Sets `sys-usb` clarifications apart into their own section to mitigate confusion
- Updates description of `config.json` file to match its expected contents
- Replaces the recommendation to skip creating default VMs with a step to rename/delete VMs with conflicting names
- Adds the recommendation to avoid using a USB keyboard (if at all possible) to the hardware section
- Updates to `Qubes OS` name to be consistent

## Status

Ready for review

### DIFF
--- a/docs/admin/backup.rst
+++ b/docs/admin/backup.rst
@@ -3,7 +3,7 @@ Backup and Restore
 
 .. include:: ../includes/top-warning.rst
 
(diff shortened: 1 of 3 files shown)


rocodes (approved):
Thank you @eloquence :)

PR #103, opened 2022-03-15
Increased CI linkcheck timeout to 30min

zenmonkeykstop:
Nightlies on this repo are failing more often than not, always about 2/3 of the way through the linkcheck command. 

This PR increases the no_output_timeout for this command from 10m to 30m.

PR #120, opened 2022-08-24
Use standard docs CI configuration

eloquence:
This uses the Debian Bullseye image to mirror typical developer configurations, and is identical to the `securedrop-docs` configuration.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,35 +1,47 @@
 version: 2
 
+common-steps:
+  - &install-deps
+    run:
+      name: Install system-level and Python dependencies
+      command: |
+        apt-get update && apt-get install -y --no-install-recommends make python3 python3-venv python3-setuptools
+        python3 -m venv .venv
+        ./.venv/bin/pip install --require-hashes -r requirements/requirements.txt
+
 jobs:
   lint:
     docker:
-      - image: cimg/python:3.10.0
+      # We use a standard Debian image to mirror a typical developer environment.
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,35 +1,47 @@
...
     steps:
       - checkout
+      - *install-deps
       - run:
-          name: Install dependencies
-          command: pip install --require-hashes -r requirements/requirements.txt
-      - run:
-          name: Run link checker
-          command: make docs-linkcheck
-          no_output_timeout: 30m
```
eloquence, line 26: This timeout was increased in https://github.com/freedomofpress/securedrop-workstation-docs/pull/103 uniquely in this repo due to timeouts on nightly checks; I'm happy to keep an eye on things and we can re-introduce it if it's still warranted.

legoktm (approved):

.circleci/config.yml
eloquence: Turns out it's indeed [failing again](https://app.circleci.com/pipelines/github/freedomofpress/securedrop-workstation-docs/696/workflows/b4b2d5d1-a66f-4360-abc1-859de5517217/jobs/680
) with `context deadline exceeded`, although it seems to run fine and quickly locally. :thinking:  Seems like it's hanging in CI on the first Whonix URL it encounters: https://www.whonix.org/wiki/Anon_Connection_Wizard

I'll try tweaking the Sphinx linkcheck settings before reintroducing the "no output" timeout in CI.

PR #121, opened 2022-08-29
Set linkcheck timeout; update Lenovo link

eloquence:
The Lenovo URL was timing out in CI (but not locally). This PR enforces a hard timeout on all linkchecks, and also updates the URL to a better location (the old URL no longer contains the actual specs).

Supersedes #103, which was reverted in #120

## Status

Ready for review. 

Note: `docs-linkcheck` only runs as a nightly job; I've tested the behavior by temporarily dropping the schedule trigger while working on this PR.  You can see a successful run here: https://app.circleci.com/pipelines/github/freedomofpress/securedrop-workstation-docs/699/workflows/a32bacc0-9939-4b69-be3b-fe28e9835274/jobs/686

### DIFF
--- a/docs/admin/hardware.rst
+++ b/docs/admin/hardware.rst
@@ -200,7 +200,7 @@ The instructions below assume the use of a Linux-based computer for the creation
 
 USB-C ports
 ~~~~~~~~~~~
-If you intend to use USB-C ports, please note that our recommended BIOS settings will disable dual USB-C/Thunderbolt ports (recognizable by the Thunderbolt logo next to the port). The T480, for example, includes two USB-C ports, `specified <https://www.lenovo.com/us/en/laptops/thinkpad/thinkpad-t-series/ThinkPad-T480/p/22TP2TT4800>`__ as follows:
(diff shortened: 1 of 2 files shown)


cfm (changes requested):
Changed link checks out.

[Here](https://github.com/freedomofpress/securedrop-workstation-docs/blob/20c240ba9a60a8c66375eeb6a46a762efd77e16a/docs/conf.py#L239) the [`linkcheck_timeout`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-linkcheck_timeout) value is given in seconds, while the `no_output_timeout` set in 5c7dad7c179a85f90a9af43327c1c9e7ee9c58ab and reverted in a4e693e140b30c92f83c7e0a8784f41ab2dd484f was given in minutes.  Is that as you intend?

eloquence:
Yes, thanks for checking - that's intentional. I figure if an individual link takes more than 30 second to load, we can treat it as a failure for the purposes of the check. The original timeout was at the CI job level, and therefore quite conservative.

cfm (approved):
Thanks for clarifying, @eloquence!

PR #109, opened 2022-06-07
document Qubes 4.1 compatibility with 2nd-gen ThinkPad T14

cfm:
Updates ThinkPad T14 hardware recommendations to reflect the 2nd-generation model compatible with Qubes 4.1, including how to install an `/rw/config/rc.local` script in `sys-net` to enable the Intel AX210 Wi-Fi controller.

## Checklist

Please do review this pull request locally via `make docs` and confirm that:

- [ ] The instructions at `admin/hardware.html` are sensible.
- The script in `includes/iwlwifi/rc.local`:
    - [ ] looks good on visual review;
    - [ ] is available (`s/main/thinkpad-t14-gen2`) at <https://raw.githubusercontent.com/freedomofpress/securedrop-workstation-docs/thinkpad-t14-gen2/docs/includes/iwlwifi/rc.local>; and
    - [ ] is embedded correctly in `admin/hardware.html`.

PR #122, opened 2022-09-26
remove `iwlwifi` workaround for Intel AX210/211 Wi-Fi chipset

cfm:
In #109, we documented how to add an `/rw/config/rc.local` script in `sys-net` to load the last known-good firmware for the AX210/211 Wi-Fi chipset.  [This workaround is no longer required with kernel 5.15.64](https://forum.qubes-os.org/t/ax210-211-wifi-card-not-showing-any-networks/9498/19), so let's remove it and just instruct users to update their systems fully.

The documentation change is straightforward, but I'd appreciate a sign-off from @maeve-fpf or someone else who can second my experience that all's well after removing our workaround on a fully-updated ThinkPad T14 with this chipset.


maeve-fpf:
I will check this out!

eloquence:
@maeve-fpf Putting this back on your radar - do you still want to double-check whether this fix is still required, or would you prefer someone else on the team take this on? I believe @nathandyer has a Gen2 T14 as well so would also be able to confirm.

nathandyer:
I just did a clean install of Qubes 4.1.2, and can confirm that the workaround I had to use before is no longer necessary. The WiFi drivers are picked up automatically with the newer kernel version (I have a T14 Gen2 with the Intel AX211 chipset).

Since everything happens automatically, I'm not sure the note here is even necessary, as no connection to a wired network is ever required, and no additional updates need to take place before the wireless adapter is usable.

I think we can omit this entirely now, but curious to hear others' thoughts.

cfm:
Thanks for confirming that happy news, @nathandyer!  I know it's my pull request and all, but I don't see any reason you can't go ahead and merge.  :-)  I've rebased from `main` without changes.

cfm:
Ah; thanks for clarifying, @nathandyer.  I've tacked that on in 90883ed.  Feel free to squash as you see fit!

nathandyer (approved):
Thanks @cfm, looks great to me!

PR #127, opened 2023-01-03
Test build step in CI

legoktm:
CI now has a build job that will build the deployment docker image just as it does in the real deployment step. Instead of forcing it to always build "main", allow a different branch to be tested with `--build-arg GIT_BRANCH=...`.

This is the same as <https://github.com/freedomofpress/securedrop-dev-docs/commit/e81a90ddf07a98e494a589c1d08564f5f824c887>, where a change broke building of the PDF, causing deployment to fail.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,10 +36,23 @@ jobs:
             source .venv/bin/activate
             make docs-linkcheck
 
+  build:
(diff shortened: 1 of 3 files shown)


nathandyer (approved):
LGTM

PR #130, opened 2023-03-24
Update Qubes version in install instructions to 4.1.2

zenmonkeykstop:
Updates the install docs to reference Qubes 4.1.2 (links to download page and ISO filename references) as Qubes 4.1.1 is no longer available for download.

## testing:
- [ ] CI is passing 
- [ ] `make  linkcheck` passes locally
- [ ] There are no references to Qubes 4.1.1 left
- [ ] The referenced ISO exist and (for extra credit) validation passes following the given instructions.

### DIFF
--- a/docs/admin/install.rst
+++ b/docs/admin/install.rst
@@ -89,7 +89,7 @@ If the Qubes hardware compatibility list entry for your computer recommends the 
 
 Download and verify Qubes OS
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


nathandyer (approved):
Results:

- [x] CI is passing 
- [x] `make linkcheck` passes locally
- [x] There are no references to Qubes 4.1.1 left
- [x] The ISO referenced in the dd command exists, and signature verification instructions (from upstream Qubes) check out

On a marginally related note, I think we might want to avoid recommending the use of dd to create install disks (it's SO easy to pick the wrong disk and overwrite something important, like your OS root, and lacks features like error correction and verification that tools like GNOME Disks or Disk Utility provide), but as Alton Brown says, "that's another show."

Approving this PR.

PR #131, opened 2023-03-24
Update mounted file paths and add warning to use Fedora as default VM Template

nathandyer:
This PR changes the path of the mounted Persistent storage drives used to copy keys and configurations from existing SVS and Journalist Workstations.

This also adds a warning to use Fedora as the default VM template, rather than Debian (I learned this one the hard way).

## Testing:
- [x] CI passes 
- [x] `make linkcheck` passes locally
- [x] Visual check on local build
- [ ] [Bonus points] Plugging in an Admin (or Journalist) Workstation drive to a Qubes 4.1.2 system, opening the Vault Qube, and verifying that the mount point is now `/media` rather than `/run/media` after entering the passphrase.

### DIFF
--- a/docs/admin/install.rst
+++ b/docs/admin/install.rst
@@ -124,6 +124,8 @@ You will be prompted to enter the FDE passphrase set during installation.
 
 After the disk is unlocked and Qubes starts, you will be prompted to complete the initial setup. Click the Qubes OS icon.
 
+.. warning:: Make sure **Fedora** is selected as the default VM template. Otherwise, you will need to manually switch the *Work* Qube template to Fedora in order to download the SecureDrop Workstation `dom0` configuration needed to begin the install.
+
 On the configuration screen, ensure that the following options are checked:
 
  - "Create default system qubes (sys-net, sys-firewall, default DispVM)"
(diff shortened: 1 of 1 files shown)


docs/admin/install.rst
```diff
@@ -124,6 +124,8 @@ You will be prompted to enter the FDE passphrase set during installation.
 
 After the disk is unlocked and Qubes starts, you will be prompted to complete the initial setup. Click the Qubes OS icon.
 
+.. warning:: Make sure **Fedora** is selected as the default VM template. Otherwise, you will need to manually switch the *Work* Qube template to Fedora in order to download the SecureDrop Workstation `dom0` configuration needed to begin the install.
```
legoktm, line 127: Does this need to be a warning vs just a bullet below that says "default VM template: Fedora" (idk what it looks like actually)? I'm thinking that just letting people follow the instructions will be easier than giving them a warning here...

docs/admin/install.rst
```diff
@@ -207,14 +209,14 @@ In order to decrypt submissions, your SecureDrop Workstation will need a copy of
   .. code-block:: sh
 
     qvm-run --pass-io vault \
-      "gpg --homedir /run/media/user/TailsData/gnupg -K --fingerprint"
```
legoktm, line 210: I plugged my Admin Workstation drive into a Fedora 37 VM and it mounted in `/run/media/...` (the old paths). Do I need to do something different? When did the paths change?

legoktm (changes requested):

nathandyer:
Closing this PR for now, the path change ended up being related to the mistake I made in changing the default template. And after reviewing our documentation again, I think it's already clear that users should just go with defaults. Not sure why I decided to venture off the beaten path during this reinstall session. Sometimes a little knowledge is a dangerous thing :) 

I have since re-installed a couple times, and the current instructions are still valid.

PR #132, opened 2023-03-30
Add a Dark Theme

nathandyer:
Now that the main SecureDrop docs have had a responsive dark theme for a while in a stable release, and most of the issues have been worked out, I think it makes sense to bring this change to our other documentation portals.

This PR uses the same style sheet as the one for docs.securedrop.org, and works exactly the same (this just makes the experience consistent for workstation users).

## Testing:
* [ ] CI passes
* [ ] Builds locally
* [ ] It dynamically changes between light and dark mode, depending on browser color preference toggle


eloquence (approved):
LGTM!

PR #133, opened 2023-03-31
Add update troubleshooting guide

eloquence:
First draft. Potential improvements:

- Screenshots (e.g. at least of an update failure)
- Common issues users may see in interactive prompts
- How to obtain logs related to individual update operations

Other considerations:

- Should we recommend running the Qubes built-in updater to attempt to resolve issues? Interactive updates generally seem to give the user more flexibility.
- Should we ask the admin to examine `launcher-detail.log`? I've erred on the side of sending it to FPF for review, given that this category of error is difficult to narrow down without domain knowledge.

### DIFF
--- a/docs/admin/troubleshooting_updates.rst
+++ b/docs/admin/troubleshooting_updates.rst
@@ -1,0 +1,172 @@
+Troubleshooting system updates
+==============================
+
+.. include:: ../includes/top-warning.rst
+
+After you log into Qubes, the SecureDrop Workstation
+preflight updater will prompt you to check for available
+system updates at least once per day.
+
+If updates fail for any reason, the preflight updater will
+not launch the SecureDrop Client application until the
+underlying issue has been resolved. This is to ensure
+that the system is in a secure state before you
+interact with SecureDrop.
+
+This guide offers troubleshooting steps for common
+update issues.
+
+Step 1: Locate the updater log
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The preflight updater runs in the ``dom0`` domain. It
+writes its log to ``~/.securedrop_launcher/logs/launcher.log``.
+Log files are rotated hourly; if you have started the updater
+again since the error occurred, you may need to check the
+previous log file.
+
+In order to examine the most recent log file:
+
+1. Open a terminal in ``dom0`` by clicking the Qubes menu
+   icon in the upper left corner of your screen, and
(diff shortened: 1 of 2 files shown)


nathandyer:
I've looked through the draft, thanks @eloquence for taking the time to write this up! I've already learned a ton. I'm holding off on any nitpicks or detailed reviews on this PR for now until it's out of draft form.

Please let me know when/if you'd like to have a chat to discuss this in more detail, especially those questions under "Other considerations." Just to note my thoughts so far:

* I agree about the interactive updates giving users more flexibility, and would prefer to go that route when possible.
* I would personally be in favor of suggesting for Admins to review `launcher-detail.log` independently if they are comfortable doing so, with the option of sending those to us for support if they wish. Although it's perhaps a bit aspirational, I'm hoping that the documentation eventually reaches a place where an administrator could potentially install, maintain, and troubleshoot a SD/SDW installation entirely independently if they don't want to involve us in any way (since we are a third-party that they may not inherently trust). It is difficult to narrow the issue down without domain knowledge, but in a perfect world that knowledge should be able to be passed to others through the documentation (again, I know, aspirational)

eloquence:
@nathandyer Thanks much for adding the screenshot. If the rest looks good to you I would suggest we get it in, and then iteratively improve it, so we can start utilizing it for support inquiries.

docs/admin/troubleshooting_updates.rst
```diff
@@ -0,0 +1,183 @@
...
+   ``sudo apt upgrade``
+
+  The SecureDrop and Whonix templates are based on Debian
+  GNU/Linux. The ``apt update`` comand will ensure the package
+  index is up-to-date, and the ``apt upgrade`` comand will
+  apply updates.
+
+3. Follow the prompts to resolve any issues. If you are
+   unsure on how to resolve an error, please contact us
+   or assistance.
```
nathandyer, line 132: Just leaving a note to remind myself about a small typo to fix

nathandyer:
Thanks @eloquence, I agree that it's best to go ahead and land this, then iterate as it makes sense. One change we might want to make is linking out to our documentation on contacting support, since we mention contacting us at a few points. That said, I think it best to wait on that for now until the dust settles there and we know if we are going to continue forward with the separate support documentation, or bundle it in with the main SD docs. 

I pushed a small commit to correct a typo; otherwise, it looked great to me! Going to go ahead and approve/merge.

PR #181, opened 2024-01-24
Document the process of increasing sync timeouts

nathandyer:
When the client loads, it tries to synchronize with the server, which may time out if there are a substantial number of sources.

This commit adds a section to the connection troubleshooting guide to advise on how to increase the timeout value.

## Testing
- [ ] Visual review
- [ ] CI passes

Note: tested in browser with `make docs` and checked with `make docs-lint`

### DIFF
--- a/docs/admin/troubleshooting_connection.rst
+++ b/docs/admin/troubleshooting_connection.rst
@@ -203,6 +203,39 @@ at the same time. Start ``sd-proxy``, which will bring up ``sd-whonix``.
 
 If this does not resolve the issue, please contact us for assistance.
 
+Customizing Synchronization Timeouts
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The SecureDrop Workstation Client application performs a synchronization on
+launch, which may time out in situations where there are a large number of 
+sources on the server. To the extent possible, we recommend regularly removing
+sources that are no longer actively engaging with you, or any sources that
+appear to be spam.
+
+If you are able to login to the Client application, but sources are not
+appearing, you may be reaching a timeout for the synchronization process.
+To temporarily increase the timeout, and give the system more time to
+finish synchronizing with the server, you can perform the following steps:
+
+1. Log into the Qubes workstation
+2. Start a system Terminal in ``dom0`` by going to the Qubes Menu, then choose
+   *Terminal Emulator*
+3. Run the following commands::
+   
+       qvm-service --enable sd-app SDEXTENDEDTIMEOUT_600
+       qvm-shutdown sd-app && sleep 5 && qvm-start sd-app
+   
+4. Start the SecureDrop Client application, like normal
+
+These instructions will change the sync timeout to 10 minutes. If the sync
+fails to complete in that time, repeat the instructions, replacing the "600"
+with a larger value. We recommend increasing the number in increments
+of 100, and checking after each change to find the best value.
+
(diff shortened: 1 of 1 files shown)


docs/admin/troubleshooting_connection.rst
```diff
@@ -203,6 +203,39 @@ at the same time. Start ``sd-proxy``, which will bring up ``sd-whonix``.
...
+
+1. Log into the Qubes workstation
+2. Start a system Terminal by going to the Qubes Menu, then choose
+   *Terminal Emulator*
+3. Run the following commands::
+   
+       qvm-service --enable sd-app SDEXTENDEDTIMEOUT_600
+       qvm-shutdown sd-app && sleep 5 && qvm-start sd-app
+   
+4. Start the SecureDrop Client application, like normal
```
cfm, line 228: This was the fastest QA procedure in freedomofpress/securedrop-client#1552, but it can also be done without resorting to a terminal, from the "Services" tab of `sd-app`'s "Qube Settings".

nathandyer: Thanks for pointing out this potential alternate workflow, @cfm! After looking into this and testing out the Services tab this afternoon, I am still in favor of using the Terminal-based approach, for a few reasons:

* If we decide to merge https://github.com/freedomofpress/securedrop-workstation/pull/857, then `sd-app` (and corresponding Qubes Settings menu) is not directly available to the user
* I found the process of adding the service to be non-intuitive (you have to click the add button, and are just presented with a blank text box, so it's not entirely obvious that that's where "SDEXTENDEDTIMEOUT_600" needs to be added at first)
* Terminal commands are suggested in several other spots within that troubleshooting page in the documentation already (for verifying connectivity and restarting Tor, mainly)

If there's anything I've overlooked, or if others feel differently, I'm certainly not _opposed_ to documenting the GUI-based approach instead, but I do feel the current approach of enabling the service via Terminal is the most straightforward and accessible.

cfm: Works for me, @nathandyer.  Thanks for looking into it.

docs/admin/troubleshooting_connection.rst
```diff
@@ -203,6 +203,39 @@ at the same time. Start ``sd-proxy``, which will bring up ``sd-whonix``.
...
+
+These instructions will change the sync timeout to 10 minutes. If the sync
+fails to complete in that time, repeat the instructions, replacing the "600"
+with a larger value. We recommend increasing the number in increments
+of 100, and checking after each change to find the best value.
+
+Once the synchronization has completed and the sources have loaded, we advise
+going through and removing any unnecessary sources.
```
legoktm, line 236: Do we want people to remove the timeout after this? Or just leave it in place?

legoktm (approved):
LGTM, one optional comment.

docs/admin/troubleshooting_connection.rst
cfm: My intuition is that folks who discover they need this should keep it for the time being (i.e., until synchronization itself is more performant).

nathandyer: I agree with @cfm, until those performance changes land I imagine this setting would be semi-permanent (with the possibility that folks might need to _increase_ the value as time goes on, assuming more sources are stored on the server).

legoktm: OK!

PR #191, opened 2024-02-21
Document the process of replacing an expired signing key

nathandyer:
This PR adds the necessary steps to the update troubleshooting guide to let folks diagnose and correct an expired Release Signing Key.

It also fixes a random typo that I spotted while making this addition.

## Testing:
- [ ] Visual review
- [ ] CI passes

### DIFF
--- a/docs/admin/troubleshooting_updates.rst
+++ b/docs/admin/troubleshooting_updates.rst
@@ -104,8 +104,61 @@ Note that ``dom0`` and ``apply_dom0`` are separate steps.
    for assistance.
 
 4. Reboot the system. ``dom0`` updates are often
-   security-sensitive, and may requrie a reboot to take
+   security-sensitive, and may require a reboot to take
    effect.
+   
+Expired SecureDrop Signing Key
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If the update fails after running ``sudo qubes-dom0-update`` as described
+above, and the terminal console displays the following message:
+
+.. code-block:: sh
+   
+   1. Certificiate 188EDD3B7B22E6A3 invalid: certificate is not alive
+       because: The primary key is not live
+       because: Expired on 2023-07-04T10:52:20Z
+   2. Key 188EDD3B7B22E6A3 invalid: key is not alive
+       because: The primary key is not live
+       because: Expired on 2023-07-04T10:52:20Z
+   [...]
+   Error: GPG check FAILED
+
+your system is trying to use an old copy of the SecureDrop Release
+Signing Key. The new, valid key will already be locally available on your
+system, so you can perform the following steps to remove the expired key
+and enable this updated key:
+
+1. Open a terminal in ``dom0`` by clicking the Qubes menu
+   icon in the upper left corner of your screen, and
+   selecting **Terminal Emulator**.
+   
+2. Run the following command:
+
+   .. code-block:: sh
+   
+      sudo rpm -q gpg-pubkey --qf '%{NAME}-%{VERSION}-%{RELEASE}\t%{SUMMARY}\n' | grep SecureDrop
+    
+   The output should look similar to:
+   
+   .. code-block:: sh
+   
+      gpg-pubkey-xxxxx-xxxxx        SecureDrop Release Signing Key <securedrop-release-key-2021@freedom.press  public key
+    
+3. Make note of the KEY ID (in the format ``gpg-pubkey-xxxxx-xxxxx``).
+
+4. Run the following commands:
+
+   .. code-block:: sh
+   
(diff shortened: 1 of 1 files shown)


docs/admin/troubleshooting_updates.rst
```diff
@@ -104,8 +104,50 @@ Note that ``dom0`` and ``apply_dom0`` are separate steps.
...
 
 4. Reboot the system. ``dom0`` updates are often
-   security-sensitive, and may requrie a reboot to take
+   security-sensitive, and may require a reboot to take
    effect.
+   
+GPG check failures
+^^^^^^^^^^^^^^^^^^
+
+If at any point you see a message that the GPG check failed,
```
zenmonkeykstop, line 113: This is a bit vague, it should include the exact error text encountered with the RPM bug and the scenario in which users are likely. Otherwise it's possible that folks will follow said instructions for any key error, which we don't want, as it may be a legitimate failure (due to breakage or malicious behaviour) which we should find out about.

zenmonkeykstop (changes requested):
More specificity around when you would follow the gpg key refresh step is required IMO. The default behaviour on any key errors should not be to just delete and reinstall, there may be valid cases where more investigation is required.

docs/admin/troubleshooting_updates.rst
```diff
@@ -104,8 +104,50 @@ Note that ``dom0`` and ``apply_dom0`` are separate steps.
...
+and you see an indication that a key has expired, your system
+may be trying to use an old copy of the SecureDrop Release Signing
+Key. 
+
```
rocodes, line 117: +1 Kev's suggestion about being specific. I would also suggest replacing the line that starts with "your system may be....." and just say "If you encounter <specific scenario>, reimport the SecureDrop Release Signing Key, which is locally available on your machine, by following the steps below."  or something. 

reasoning:
- "may be trying to" makes it sound like we don't know the root cause, but we do, or that the behaviour is inconsistent
- mentioning that the good key is available already on their machine in order to clarify that we aren't instructing sometimes users perform some workaround, and that the security of their system is not in doubt.

docs/admin/troubleshooting_updates.rst
```diff
@@ -104,8 +104,50 @@ Note that ``dom0`` and ``apply_dom0`` are separate steps.
...
+To remove the expired key and replace it with the new, valid key, you
+can perform the following steps:
+
```
rocodes, line 120: (see above, and I would remove "replace it with the new valid key" in favour of wording above or similar)

docs/admin/troubleshooting_updates.rst
nathandyer: Thanks @zenmonkeykstop! I just made it a bit more explicit, please let me know if you think there's anything more we can add it improve it.

docs/admin/troubleshooting_updates.rst
nathandyer: That's a great suggestion, I just added a commit to implement your suggestion. Please let me know if you think it's sufficient, or if you'd rather go with your wording verbatim.

docs/admin/troubleshooting_updates.rst
```diff
@@ -107,16 +107,15 @@ Note that ``dom0`` and ``apply_dom0`` are separate steps.
...
-
-If at any point you see a message that the GPG check failed,
-and you see an indication that a key has expired, your system
-may be trying to use an old copy of the SecureDrop Release Signing
-Key. 
-
-To remove the expired key and replace it with the new, valid key, you
-can perform the following steps:
+Expired SecureDrop Signing Key (2021)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
rocodes, line 111: I hate to do this to you, but I don't think (2021) should be in the title. The new master key fingerprint was generated in 2021, but the specific pubkey was valid 2022-july to 2023- july. I think "Expired SecureDrop Signing Key" is fine.

In the next paragraph I'd start with

"If after running `sudo qubes-dom0-update` as described above, the update fails and the terminal console displays the following message:
```
1. Certificiate [sic] 188EDD3B7B22E6A3 invalid: certificate is not alive
    because: The primary key is not live
    because: Expired on 2023-07-04T10:52:20Z
2. Key 188EDD3B7B22E6A3 invalid: key is not alive
    because: The primary key is not live
    because: Expired on 2023-07-04T10:52:20Z
[...]
Error: GPG check FAILED
```
the system is trying to use an old copy.... (etc)

rocodes: and yes there is a typo in the console output ("certificiate") and we should leave the typo in the docs

zenmonkeykstop (approved):

PR #196, opened 2024-03-18
Document VeraCrypt support and drive creation process

legoktm:


docs/admin/provisioning_usb.rst
```diff
@@ -38,5 +39,45 @@ of its airgap:
...
+
+Creating a VeraCrypt-encrypted drive
+------------------------------------
+
+- If it isn't already done, download and install the `VeraCrypt software <https://www.veracrypt.fr/en/Home.html>`_.
+- Start VeraCrypt from your computer's application or software interface.
+- Click **Create Volume**
+- Select **Encrypt a non-system partition/drive** and click **Next**.
+- Select **Standard VeraCrypt volume** and click **Next**
+- Connect your fresh USB stick and click **Select Device...** to choose your USB.
```
legoktm, line 53: Do we need to provide more details on how to select the correct drive? Not sure what the interface looks like on macOS/Windows

docs/admin/provisioning_usb.rst
```diff
@@ -38,5 +39,45 @@ of its airgap:
...
+
+  - You may see a warning that says "We strongly recommend that inexperienced
```
legoktm, line 55: Do we want people to only encrypt partitions? Means we need to guide them into creating a partition first I think

docs/admin/provisioning_usb.rst
```diff
@@ -38,5 +39,45 @@ of its airgap:
...
+  Make sure that you select the correct device!
+
+- You will be prompted to set a password. This password
+  should be strong - a 6-word `Diceware <https://en.wikipedia.org/wiki/Diceware>`_
+  passphrase is highly recommended.
+- You will be asked if you need to store large files, select **No** and click **Next**.
+- Select the following options in the Volume Format dialog:
+
+  - Filesystem: FAT
+  - Quick Format: unselected
```
legoktm, line 70: Should we enable Quick Format?

legoktm (commented):
Some questions as I was writing this. Also since people need to unlock the drive before it can be used, do we need documentation on how to use e.g. GNOME Disks to do that?

docs/admin/provisioning_usb.rst
rocodes: I think/hope we can leave that to people to figure out

docs/admin/provisioning_usb.rst
rocodes: They can encrypted a whole device or a partition, either way

docs/admin/provisioning_usb.rst
rocodes: I haven't tested with it enabled. The only other things I would add here are:
- leave the default PIM settings
- Don't create a hidden VeraCrypt partition

rocodes:
> since people need to unlock the drive before it can be used, do we need documentation on how to use e.g. GNOME Disks to do that?

I guess yeah that's a good idea. Here are some screenshots, I can tack on some copy tomorrow if you like. The gist is:

- Open `nautilus` in `sd-devices` (Insert USB, and wait for `sd-devices` VM to start. Then from Q application menu -> sd-devices -> "Files")
- There should be an entry called "N GB Possibly Encrypted"
- Double-click on it, and then input passphrase and/or other settings as needed

Here are some screenshots. I also have not tested the "Windows system" checkbox :upside_down_face: 

veracrypt

veracrypt_unlock

docs/admin/provisioning_usb.rst
rocodes: No, they should leave it unselected

docs/admin/provisioning_usb.rst
legoktm: Oh good point, I've added something like "We disagree with this recommendation..." just to make it clearer it can be ignored.

docs/journalist/submissions.rst
```diff
@@ -56,26 +56,33 @@ administrator. The printer must be plugged into the computer's USB port.
...
+   1. Open the file menu by clicking on the **Q** application menu (in the top left),
+      select **sd-devices** and click **Files**.
+   2. In the left sidebar, there should be an entry labeled **# GB Possibly Encrypted**,
+      click it.
+      |screenshot_veracrypt_sd_devices_files|
+   3. You will be prompted for the password configured for this USB drive:
+      - Volume type: leave both unchecked
+      - PIM: leave empty
+      - Password: drive's password
+      - Forget password immediately: selected
```
rocodes, line 73: this is a nit, but: rst jumbles up this formatting a bit.  Also, the PIM and Hidden Volume, it depends how they set it up. If they changed the PIM when configuring the drive, they need to put in their custom number here.  And in theory if they unlock the hidden volume here we can probably export to it. So I wasn't sure whether to put our suggestions or just put "Settings consistent with original configuration" or something

Screenshot 2024-03-19 at 13-30-08 Working with submissions — SecureDrop Workstation 0 0 1 documentation

legoktm: > this is a nit, but: rst jumbles up this formatting a bit.

oops, I was so focused on the images I missed this. fixed.

>  Also, the PIM and Hidden Volume, it depends how they set it up. If they changed the PIM when configuring the drive, they need to put in their custom number here

I was thinking that this should match the drive creation steps (which now I realize should be more specific about leaving PIM, etc. blank). We could also have it be like `PIM: leave empty (unless you set one during creation)`?

legoktm: Alternatively:
```patch
diff --git a/docs/journalist/submissions.rst b/docs/journalist/submissions.rst
index ec93355..b9ca79e 100644
--- a/docs/journalist/submissions.rst
+++ b/docs/journalist/submissions.rst
@@ -66,15 +66,11 @@ Currently, a LUKS- or VeraCrypt-encrypted USB drive is required for exporting su
    2. In the left sidebar, there should be an entry labeled **# GB Possibly Encrypted**,
       click it.
       |screenshot_veracrypt_sd_devices_files|
-   3. You will be prompted for the password configured for this USB drive:
-
-      - Volume type: leave both unchecked
-      - PIM: leave empty
-      - Password: drive's password
-      - Forget password immediately: selected
-
+   3. You will be prompted for the password configured for this USB drive. In most cases
+      you will just need to enter the drive's password and leave all the other fields blank
+      unless you picked different settings during provisioning. Make sure you select
+      **Forget password immediately** and then click **Connect**.
       |screenshot_veracrypt_sd_devices_files_unlock|
-   4. Click **Connect**.
 
 3. Back in your source's conversation, click **Export**.
    |screenshot_export_dialog|
```

legoktm: And maybe have the `select **Forget password immediately** and then click **Connect**` be step 4 so it doesn't get overlooked.

rocodes: (per call, resolving - lgtm and I agree it's simpler to leave as is in case one person provisions, one person uses the drive)

rocodes (approved):
- [x] visual review
- [x] `make docs`,  `make docs-lint`, `make docs-linkcheck`
- [x] provisioning steps are correct 

thanks @legoktm  :)

PR #217, opened 2024-06-17
Break up the install guide into a dedicated section

### LINKED ISSUE #212: Break up the "Installing and Configuring SecureDrop Workstation" article
We currently have a single document which outlines the entire installation process, start to finish. It includes subsections for pre-install tasks (like installing Qubes), configuring Qubes, installing SecureDrop Workstation, and even has a section on troubleshooting. This makes for a very, very long document that is easy to get lost in during the install process.

Much like the Server installation guide, we should break this up into multiple articles, with a "Next" button at the end of each significant section to help walk an admin through the installation process.

nathandyer:
## Description

This PR is a slight reorganization of the Admin Guide into two sections, similar to how SecureDrop (server) docs are done. With these changes, there is now one section dedicated to the installation, and one section to general reference for Admins.

Along with the changes, the individual pages, screenshots, and images are reorganized, and internal references have been updated where necessary.

## Testing
- [ ] Visual
- [ ] CI passes


eloquence:
Took a spin, looks good to me, and I like that it now more closely mirrors the SecureDrop server doc structure. Will let it sit for a bit in case others have thoughts.

eloquence (approved):
LGTM! Per chat with Kev, landing this now so the structure is in place; we can always follow up as needed.

Issue #221, opened 2024-06-28
Update docs for Qubes 4.2


eloquence:
First round of issues I've noticed:
- [x] download URLs need to be updated to point to 4.2
- [ ] menu references need to be updated per the new menu GUI
- [x] explanations for Qubes OS updater need to be .. updated  per the new GUI

PR #222, opened 2024-06-28
Qubes 4.2 download, menu and updater changes

eloquence:
- Point to Qubes 4.2.1 ISO
- Update menu refrences consistent with the new Q menu
- Describe updater consistent with its new GUI

The new Q menu is fairly complex, and describing it in text alone is tricky. To recap:
- There's a vertical section with icons for search, domains, favorites, and system
- For domains, there's a horizontal sub-section "APPS |  TEMPLATES | SERVICE". 
- The menu always opens in the domains section
- The selection of "APPS", "TEMPLATES" or "SERVICE" is persistent when you open and close the menu

I've therefore adopted the following conventions:
- When the user needs to switch to the system section, indicate this with ⚙️
- When the user needs to operate on domains, always indicate whether they need to be in the APPS, TEMPLATES or SERVICE sub-section, because the menu may be on a different sub-section when they open it
- I'm not indicating explicitly when the user needs to be in the "Domains" section, since the menu always opens on that section

Refs #221

### DIFF
--- a/docs/admin/install/install.rst
+++ b/docs/admin/install/install.rst
@@ -9,7 +9,7 @@ In order to decrypt submissions, your SecureDrop Workstation will need a copy of
 
 - First, use the network manager widget in the upper right panel to disable your network connection. These instructions refer to the ``vault`` VM, which has no network access by default, but if the SVS USB is attached to another VM by mistake, this will offer some protection against exfiltration.
 
-- Next, choose **Q > Domain: vault > vault: Files** to open the file manager in the ``vault`` VM.
+- Next, choose **Q > APPS > vault > Thunar File Manager** to open the file manager in the ``vault`` VM.
 
 - Connect the SVS USB to a USB port on the Qubes computer, then use the devices widget in the upper right panel to attach it to the ``vault`` VM. There will be three entries for the USB in the section titled **Data (Block) Devices**. Choose the *unlabeled* entry (*not* the one labeled "TAILS") annotated with a ``sys-usb`` text that ends with a number, like ``sys-usb:sdb2``. That is the persistent volume.
 
@@ -19,7 +19,7 @@ In order to decrypt submissions, your SecureDrop Workstation will need a copy of
 
   |Unlock TailsData|
 
-- Open a ``dom0`` terminal via **Q > Terminal Emulator**, and run the following command to list the SVS submission key details, including its fingerprint:
+- Open a ``dom0`` terminal via **Q > ⚙️ > Other > Xfce Terminal**, and run the following command to list the SVS submission key details, including its fingerprint:
 
   .. code-block:: sh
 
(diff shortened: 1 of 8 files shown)


eloquence:
The build error is due LaTeX stumbling over the ⚙️ character. If we want to keep it, that's probably fixable (change the branch in `deploy/Dockerfile` and debug builds with ``podman build -t sdwdocs -f deploy/Dockerfile .``); otherwise I'd suggest changing that to "Gear symbol".

legoktm:
> The build error is due LaTeX stumbling over the ⚙️ character. If we want to keep it, that's probably fixable (change the branch in `deploy/Dockerfile` and debug builds with `podman build -t sdwdocs -f deploy/Dockerfile .`); otherwise I'd suggest changing that to "Gear symbol".

Via https://github.com/sphinx-doc/sphinx/issues/10469 I found https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-latex_engine - I can experiment with that a bit before giving up on unicode.

legoktm:
Unfortunately switching to `xelatex` is only half of the solution, then we need to find a font that supports the gear icon (`⚙ (U+2699)`) in bold. Neither Noto nor DejaVu Serif have it. For some reason trying to set `\setmainfont{Noto Color Emoji}` doesn't work. Using `Symbola` does though, but I'm not super impressed with the rest of that font's quality.

```diff
diff --git a/deploy/Dockerfile b/deploy/Dockerfile
index e434507..f27947f 100644
--- a/deploy/Dockerfile
+++ b/deploy/Dockerfile
@@ -1,7 +1,8 @@
 FROM debian:bookworm AS sphinx
 
 ARG GIT_BRANCH=main
-RUN apt-get -q update && apt-get -qy upgrade && apt-get -qy install git make latexmk texlive-latex-extra python3-poetry
+RUN apt-get -q update && apt-get -qy upgrade && apt-get -qy install git make latexmk texlive-xetex texlive-latex-extra python3-poetry \
+    fonts-freefont-otf font-symbola texlive-fonts-extra-links
 COPY ./ .
 RUN poetry install
 RUN deploy/build $GIT_BRANCH
diff --git a/docs/conf.py b/docs/conf.py
index 4a59811..a105b6e 100644
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -160,7 +160,13 @@ htmlhelp_basename = 'SecureDropWorkstationdoc'
 
 # -- Options for LaTeX output ---------------------------------------------
 
+latex_engine = 'xelatex'
 latex_elements = {
+    'fontpkg': r'''
+\setmainfont{DejaVu Serif}
+\setsansfont{DejaVu Sans}
+\setmonofont{DejaVu Sans Mono}
+''',
     # The paper size ('letterpaper' or 'a4paper').
     #
     # 'papersize': 'letterpaper',
@@ -177,7 +183,6 @@ latex_elements = {
     # seem to make this magically work, so define individual characters.
     'preamble': r'''
     \usepackage{amssymb}
-    \DeclareUnicodeCharacter{25B8}{$\blacktriangleright$}
     '''
 
     # Latex figure (float) alignment
```

I'm out of timebox so I'm going to say that we won't be able to use a gear emoji for the 1.0.0 release. We should aim to switch to xelatex or another Unicode supporting rendering engine though.

legoktm:
(I've copied the above thread to https://github.com/freedomofpress/securedrop-docs/issues/576 so we don't forget about this)

eloquence:
Most of this was incorporated into #228 (thanks :); I spotted a couple of menu references that still need to be updated, will open a smaller PR for those.

PR #224, opened 2024-07-02
Overhaul the hardware guide

### LINKED ISSUE #69: Recommend USB3 drives with minimum space requirements for USB BIOS install
There's not yet any mention of what sort of USB drives to get when you get to the part of docs asking you to move bios code into a USB drive: https://workstation.securedrop.org/en/stable/admin/hardware.html. Would be helpful to recommend USB3 drives with a minimum size.

Related note that the docs also call it a "USB" which is technically a protocol and not a specific kind of USB device, which probably won't cause confusion realistically but "USB drive" feels more accurate.

### LINKED ISSUE #210: Set 32GB as the minimum RAM amount for SecureDrop Workstation
Right now, our docs recommend 16GB of RAM, but 32GB of RAM for "production use." 

Moving forward, I think it makes sense to set 32GB as the standard for all uses, production or otherwise (it's unclear to me what we see as "non-production" use - setting up a demo system? For the Workstation to function a full SecureDrop server installation is needed, which I think implies production-use or soon-to-be-production use in any case).

### LINKED ISSUE #211: Phase out specific hardware recommendations
Right now, we're recommending a handful of specific models of laptops for use with Qubes and SecureDrop Workstation (currently all ThinkPads).

Given the maintenance burden, we should consider removing specific recommendations in favor of pointing users to the Qubes Hardware Compatibility List (HCL): www.qubes-os.org/hcl/

nathandyer:
This PR overhauls the hardware guide by making a few key changes:

1. It removes our specific laptop recommendations, instead referring people to the Qubes hardware compatibility list.
2. It clarifies a few specific requirements for hardware, namely 32GB of RAM (removing the confusing language about production use), as well as a USB 3.0 flash drive with minimum 8GB of capacity.
3. It provides a generalized BIOS update section, which points folks to `fwupd` on supported systems, and to their manufacturers instructions for instances where fwupd is not supported.

## Fixes

Fixes #211 
Fixes #210 
Fixes #69 

## Testing
- [ ] Visual inspection
- [ ] CI passes

### DIFF
--- a/docs/admin/install/prepare.rst
+++ b/docs/admin/install/prepare.rst
@@ -16,7 +16,7 @@ To rotate passphrases for accounts, please see the `instructions <https://docs.s
 
 Apply BIOS updates and check settings
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-Before beginning the Qubes installation, make sure that your Qubes-compatible computer's BIOS is updated to the latest available version. If you're using one of the recommended ThinkPad T-series models, see the section on :ref:`thinkpad_t_series`. The process will be different for other makes and models, and can usually be found on their respective support sites.
+Before beginning the Qubes installation, make sure that your Qubes-compatible computer's BIOS is updated to the latest available version. For more details about this process, see the section on :ref:`general_BIOS_update`.
 
 Once the BIOS is up-to-date, boot into the BIOS setup utility and update its settings. Note that not all BIOS versions will support the items listed, but if available following changes are recommended:
 
--- a/docs/admin/reference/hardware.rst
+++ b/docs/admin/reference/hardware.rst
@@ -10,108 +10,24 @@ Qubes OS hardware requirements
 In order to install and use SecureDrop Workstation, you will need a Qubes-Compatible computer with the following specifications:
 
 - 64-bit Intel or AMD processor with virtualization support
-- a minimum of 16GB RAM (32GB recommended for production use)
+- a minimum of 32GB RAM
 - sufficient disk space for the Qubes OS base install and SecureDrop Workstation VMs (a 128GB or greater SSD is recommended)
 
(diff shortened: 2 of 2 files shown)


legoktm (commented):
Few notes:

* I'm a bit uncertain on whether pointing people at Qubes HCL instead of specific requirements is a good idea. When I visit <https://www.qubes-os.org/hcl/#hardware-laptops> there are so many options listed that I still don't know what to buy and which columns are optional for SDW and which are mandatory. (I also don't remember if we already discussed this and came to a conclusion - OK to ignore me if that's the case)
* Clarifying a minimum of 32GB RAM unconditionally LGTM.
* I just learned via @deeplow that Qubes how has fwupd integration - https://github.com/freedomofpress/securedrop-workstation/issues/1125; is that stable enough for us to recommend via docs until we build it into the updater? Otherwise I think the steps you added are fine.
* I think the PCI reset instructions should live somewhere? (I just ran into it again this morning). Maybe we can just link out to https://www.qubes-os.org/doc/pci-troubleshooting/ and possibly improve that for better GUI instructions? Maybe we have a "Hardware workarounds" page that also includes https://github.com/freedomofpress/securedrop-workstation-docs/issues/226

zenmonkeykstop:
Seconded on the HCL - ISTR it was discussed and we were kindof leaning towards:
- 1x Qubes-certified recco (most likely the Novacustom NV41 (coreboot/heads still undecided)
- 1x "mainstream" recco (most likely a Lenovo T-series known to work well with 4.2)
- only as a 3rd option would we mention the HCL, with limited support implications.

nathandyer:
Thanks both for the review and discussion! I'll table this for now and we can return to it again when the timing is right.

eloquence:
Superseded by our recommendation to use Qubes-certified hardware; will open a separate smol PR for #69.

PR #228, opened 2024-07-09
Update Install Guide for 4.2

nathandyer:
This PR updates the install guide to reference all the changes users will now encounter as they step through an installation using Qubes 4.2. While most of these changes a fairly straightforward (for example, updating the locations of items inside the Qubes menu), a few of the changes I've made should be made explicit here:

- I've added a warning to the top of the prepare and install pages noting that not all packages are available in the repos, and that these instructions will not work until that happens. We should remove this once 1.0.0 is released and packages are available on yum.securedrop.org.
- Likewise, there are links to the Qubes 4.2.2-rc1 iso (mainly as a placeholder) that need to be updated to stable 4.2.2 once that is ready.
- I removed the Fedora 40 upgrade guide, and the references to it within; from here forward, no Workstation users will encounter a situation where they would need to manually upgrade to the Fedora 40 template.
- I've expanded the docs in a few small ways, such as noting that the "Unable to unlock 'TailsData'" warning is to be expected, and to make it explicit how to easily verify the Qubes ISO

## Testing
- [ ] Visual review
- [ ] CI passes


docs/admin/install/install.rst
```diff
@@ -76,9 +83,9 @@ Users of SecureDrop Workstation must enter their username, passphrase and two-fa
 
 In order to set up KeePassXC for easy use:
 
-- Add KeePassXC to the application menu by selecting it from the list of available apps in **Q > Domain: vault > vault: Qube Settings > Applications** and pressing the button labeled **>** (do not press the button labeled **>>**, which will add *all* applications to the menu).
+- Add KeePassXC to the application menu by selecting it from the list of available apps in **Q > Apps > vault > Settings > Applications** and pressing the button labeled **>** (do not press the button labeled **>>**, which will add *all* applications to the menu).
```
legoktm, line 86: This is a pre-existing issue, but "Applications" is part of a separate menu, and I think it's misleading to make it think it's part of the same Q menu.

(Not asking for a fix, just flagging for a future potential improvement)

docs/admin/install/install.rst
```diff
@@ -152,13 +159,13 @@ With the key and configuration available in ``dom0``, you're ready to set up Sec
```
legoktm, line 159: The f32 in the `[securedrop-workstation-temporary]` section needs to be switched to f37 as well.

docs/admin/install/prepare.rst
```diff
@@ -38,15 +42,21 @@ If the Qubes hardware compatibility list entry for your computer recommends the
 
 Download and verify Qubes OS
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-On the working computer, download the Qubes OS ISO for version ``4.1.2`` from `https://www.qubes-os.org/downloads/ <https://www.qubes-os.org/downloads/#qubes-release-4-1-2>`_. The ISO is 5.4 GiB approximately, and may take some time to download based on the speed of your Internet connection.
+On the working computer, download the Qubes OS ISO and cryptographic hash values for version ``4.2.2rc1`` from `https://www.qubes-os.org/downloads/ <https://www.qubes-os.org/downloads/#qubes-release-4-2-2-rc1>`_. The ISO is 6.9 GB approximately, and may take some time to download based on the speed of your Internet connection.
+
+Follow the linked instructions to `verify the ISO <https://www.qubes-os.org/security/verifying-signatures/#how-to-verify-detached-pgp-signatures-on-qubes-isos>`_. If both the ISO and cryptographic hash values are stored in the same directory, you can quickly verify by running:
```
legoktm, line 47: Instead of "If both ...are stored in the same directory" can we tell them to download them to the same directory?

"Ensure that the ISO and hash values are in the same directory, then run ..." ?

docs/admin/install/prepare.rst
```diff
@@ -38,15 +42,21 @@ If the Qubes hardware compatibility list entry for your computer recommends the
...
+
+.. code-block:: sh
+
+  gpg --keyserver-options no-self-sigs-only,no-import-clean --fetch-keys https://keys.qubes-os.org/keys/qubes-release-X-signing-key.asc
+  gpg -v --verify Qubes-R4.2.2-rc1-x86_64.iso.DIGESTS
```
legoktm, line 52: Let's show the output here which should show the fingerprint here as a confirmation they imported the correct key as well as demonstrating what good output looks like because GPG is a nightmare

docs/admin/install/prepare.rst
```diff
@@ -38,15 +42,21 @@ If the Qubes hardware compatibility list entry for your computer recommends the
...
 Download and verify Qubes OS
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-On the working computer, download the Qubes OS ISO for version ``4.1.2`` from `https://www.qubes-os.org/downloads/ <https://www.qubes-os.org/downloads/#qubes-release-4-1-2>`_. The ISO is 5.4 GiB approximately, and may take some time to download based on the speed of your Internet connection.
+On the working computer, download the Qubes OS ISO and cryptographic hash values for version ``4.2.2rc1`` from `https://www.qubes-os.org/downloads/ <https://www.qubes-os.org/downloads/#qubes-release-4-2-2-rc1>`_. The ISO is 6.9 GB approximately, and may take some time to download based on the speed of your Internet connection.
+
+Follow the linked instructions to `verify the ISO <https://www.qubes-os.org/security/verifying-signatures/#how-to-verify-detached-pgp-signatures-on-qubes-isos>`_. If both the ISO and cryptographic hash values are stored in the same directory, you can quickly verify by running:
+
+.. code-block:: sh
+
+  gpg --keyserver-options no-self-sigs-only,no-import-clean --fetch-keys https://keys.qubes-os.org/keys/qubes-release-X-signing-key.asc
```
legoktm, line 51: The X should be 4.2, i.e. https://keys.qubes-os.org/keys/qubes-release-4.2-signing-key.asc

docs/admin/install/prepare.rst
```diff
@@ -38,15 +42,21 @@ If the Qubes hardware compatibility list entry for your computer recommends the
...
+  gpg -v --verify Qubes-R4.2.2-rc1-x86_64.iso.DIGESTS
 
-Follow the linked instructions to `verify the ISO <https://www.qubes-os.org/security/verifying-signatures/#how-to-verify-detached-pgp-signatures-on-qubes-isos>`_.
 
 Once you've verified the ISO, copy it to your installation medium - for example, if using Linux and a USB stick, using the command:
```
legoktm, line 55: (We should really get them to use GNOME Disks like https://github.com/freedomofpress/securedrop-docs/issues/338 but that's for another time...)

legoktm (changes requested):
Flagged a few things. Per discussion let's change the Qubes URLs to point to a future 4.2.2 release, if it's not ready by the time we go stable we can change it down to 4.2.1 and restore the upgrade Fedora instructions.

Erik had a few more things in https://github.com/freedomofpress/securedrop-workstation-docs/pull/222, do you want to cherry-pick those here too?

docs/admin/install/prepare.rst
nathandyer: It's funny because in my notes during the installation this week I wrote "dear $deity, why are we asking people to run dd", but then decided that was probably a post-1.0.0 battle. But I'm 1000% in agreement. :)

docs/admin/install/prepare.rst
nathandyer: Good call! I also added a quick blurb about what to do if the verification fails.

docs/admin/install/prepare.rst
```diff
@@ -38,15 +42,43 @@ If the Qubes hardware compatibility list entry for your computer recommends the
...
+  gpg: textmode signature, digest algorithm SHA256, key algorithm rsa4096
+  
+Specifically, you will want to make sure that you see "Good signature" listed in the text. If it does not report a good signature, try deleting the ISO and downloading it again.
 
 Once you've verified the ISO, copy it to your installation medium - for example, if using Linux and a USB stick, using the command:
 
 .. code-block:: sh
 
-  sudo dd if=Qubes-R4.1.2-x86_64.iso of=/dev/sdX bs=1048576 && sync
+  sudo dd if=Qubes-R4.2.2-rc1-x86_64.iso of=/dev/sdX bs=1048576 && sync
```
zenmonkeykstop, line 81: Same note re: rc versions.

docs/admin/install/prepare.rst
```diff
@@ -127,13 +160,8 @@ Apply updates to system templates (estimated wait time: 45-60 minutes)
...
 
-- Once Tor has connected, select **Q > Qubes Tools > Qubes Update** to update the system VMs. in the ``[Dom0] Qubes Updater`` window, first check ``Enable updates for qubes without known available updates``, then check all entries in the list above except for dom0 (which you have already updated in the previous step). Then, click **Next**. The system's VMs will be updated sequentially - this may take some time. When the updates are complete, click **Finish**.
-
-Install Fedora 40 template
-~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-See :doc:`../reference/upgrading_fedora`.
+- Once Tor has connected, select **Q > Settings Gear > Qubes Tools > Qubes Update** to update the system VMs. in the ``[Dom0] Qubes Updater`` window, check all entries in the list above except for dom0 (which you have already updated in the previous step). Then, click **Next**. The system's VMs will be updated sequentially - this may take some time. When the updates are complete, click **Finish**.
```
zenmonkeykstop, line 167: Same note re gear icon. why are we waiting for tor to be up here tho? IIRC the install settings assume updates over public internet.

zenmonkeykstop: Oh nm, I guess the whonix template is still using Tor for updates, so it has to be up.

docs/admin/install/install.rst
```diff
@@ -2,24 +2,31 @@ Installing SecureDrop Workstation
...
 - In the the ``vault`` file manager, select **+ Other Locations**, then click the persistent volume's listing in the right panel. It will be named ``N GB encrypted``, where N is the size of the persistent volume. Enter the SVS persistent volume passphrase to unlock and mount it.
 
+  .. note::
+
+    You will receive a message that says **Failed to open directory "TailsData"**. This is normal behavior and will not cause any issues with the subsequent steps.
+
   |Unlock TailsData|
 
-- Open a ``dom0`` terminal via **Q > Terminal Emulator**, and run the following command to list the SVS submission key details, including its fingerprint:
+- Open a ``dom0`` terminal via **Q > Settings Gear > Other > Xfce Terminal**, and run the following command to list the SVS submission key details, including its fingerprint:
```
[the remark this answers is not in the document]
legoktm: I don't think we need to be that verbose, "Settings Gear" seems pretty clear to me?

docs/admin/install/prepare.rst
```diff
@@ -2,6 +2,10 @@ Pre-install Tasks
 =================
 .. include:: ../../includes/top-warning.rst
 
+.. warning::
+  The instructions will not work until the necessary packages are published in the stable SecureDrop repositories. Please try again soon, once the stable packages have been updated.
+
+
 Rotate legacy passphrases
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 To ensure that all passphrases meet the security requirements of the system, you must rotate the passphrases of any *Journalist Interface* users whose accounts were set up on or before September 12, 2017.
```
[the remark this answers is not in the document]
nathandyer: I think you're probably right. I removed it, but tacked it on as an additional commit in case someone feels we _do_ still need to leave it in.

docs/admin/install/install.rst
```diff
@@ -2,24 +2,31 @@ Installing SecureDrop Workstation
 =================================
 .. include:: ../../includes/top-warning.rst
 
+.. warning::
```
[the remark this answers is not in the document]
nathandyer: That makes sense. Since the timing is looking like the docs changes should land fairly closely with the packages, I think we are okay dropping the warning altogether. We can wait to tag the stable version until the packages are up, and if these changes are on `latest` for just a day or so I don't imagine it would cause anyone heartburn.

nathandyer:
Alright, I think I've addressed everyone's feedback. If I've overlooked anything, please let me know. A rough list of changes I made:
- Removed the warning and all language about pre-release/unstable software
- Updated Qubes references to point to 4.2.2 (not rc), with the knowledge we can fall back to 4.2.1 if need be
- Added in the additional changes from Erik's draft PR in #222 
- Added a commit to remove the section about rotating legacy passphrases (we can drop the commit if we want to leave that in for now)
- It seems there is still some debate about how to handle the "Settings gear" references. The unicode icon is not an option for this release, and I think like @legoktm I prefer the more concise option of just referring to the "Settings gear icon" within the path - but will be glad to change it to the more verbose option ("on the left sidebar"... etc.) if we prefer it.

legoktm:
I'm going to walk through this via yum-qa/apt-qa and approve if it all goes well.

docs/admin/install/prepare.rst
```diff
@@ -127,13 +153,8 @@ Apply updates to system templates (estimated wait time: 45-60 minutes)
...
 
-- Once Tor has connected, select **Q > Qubes Tools > Qubes Update** to update the system VMs. in the ``[Dom0] Qubes Updater`` window, first check ``Enable updates for qubes without known available updates``, then check all entries in the list above except for dom0 (which you have already updated in the previous step). Then, click **Next**. The system's VMs will be updated sequentially - this may take some time. When the updates are complete, click **Finish**.
-
-Install Fedora 40 template
-~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-See :doc:`../reference/upgrading_fedora`.
+- Once Tor has connected, select **Q > Settings Gear > Qubes Tools > Qubes Update** to update the system VMs. in the ``[Dom0] Qubes Updater`` window, check all entries in the list above except for ``dom0`` (which you have already updated in the previous step). Then, click **Update**. The system's VMs will be updated sequentially - this may take some time. When the updates are complete, click **Next**.
```
legoktm, line 160: Once you click "Next", it prompts you to "Finish and restart/shutdown 4 qubes" - I think we should include that final step and waiting for everything to restart.

(still working through the rest)

docs/admin/install/install.rst
```diff
@@ -9,17 +9,21 @@ In order to decrypt submissions, your SecureDrop Workstation will need a copy of
...
 
-- Next, choose **Q > Domain: vault > vault: Files** to open the file manager in the ``vault`` VM.
+- Next, choose **Q > Apps > vault > Thunar File Manager** to open the file manager in the ``vault`` VM.
 
 - Connect the SVS USB to a USB port on the Qubes computer, then use the devices widget in the upper right panel to attach it to the ``vault`` VM. There will be three entries for the USB in the section titled **Data (Block) Devices**. Choose the *unlabeled* entry (*not* the one labeled "TAILS") annotated with a ``sys-usb`` text that ends with a number, like ``sys-usb:sdb2``. That is the persistent volume.
 
   |Attach TailsData|
 
 - In the the ``vault`` file manager, select **+ Other Locations**, then click the persistent volume's listing in the right panel. It will be named ``N GB encrypted``, where N is the size of the persistent volume. Enter the SVS persistent volume passphrase to unlock and mount it.
```
legoktm, line 18: (future improvement) Give guidance on whether to choose "Forget password immediately" or "Remember password until you logout" (I think the former)

docs/admin/install/install.rst
```diff
@@ -9,17 +9,21 @@ In order to decrypt submissions, your SecureDrop Workstation will need a copy of
...
+  .. note::
+
+    You will receive a message that says **Failed to open directory "TailsData"**. This is normal behavior and will not cause any issues with the subsequent steps.
+
   |Unlock TailsData|
 
-- Open a ``dom0`` terminal via **Q > Terminal Emulator**, and run the following command to list the SVS submission key details, including its fingerprint:
+- Open a ``dom0`` terminal via **Q > Settings Gear > Other > Xfce Terminal**, and run the following command to list the SVS submission key details, including its fingerprint:
 
   .. code-block:: sh
```
legoktm, line 29: Line 49: `- In the ``vault`` file manager, select **+ Other Locations** and eject the TailsData volume, then disconnect the SVS USB.`

There's no "Other Locations" in Thunar, I think we can recommend right-clicking on the TailsData sidebar entry and clicking "Unmount"

docs/admin/install/install.rst
```diff
@@ -152,13 +156,13 @@ With the key and configuration available in ``dom0``, you're ready to set up Sec
...
 
   where ``<versionNumber>`` is the release version number you noted above. The command output should match the following text:
 
   .. code-block:: none
 
-    securedrop-workstation-dom0-config-<versionNumber>-1.fc32.noarch.rpm:
+    securedrop-workstation-dom0-config-<versionNumber>-1.fc37.noarch.rpm:
       Header V4 RSA/SHA512 Signature, key ID 7b22e6a3: OK
       Header SHA256 digest: OK
       Header SHA1 digest: OK
```
legoktm, line 168: I don't see the `  V4 RSA/SHA512 Signature, key ID 7b22e6a3: OK` line (right above MD5 digest) anymore when I run this in the Fedora 40-based VM.

legoktm (changes requested):
OK, finished going through, just about everything is good, I flagged one of the "Other locations" needing to be updated for Thunar, there are a few more places if you search in the file for it.

docs/admin/install/install.rst
nathandyer: No time like the present :)

nathandyer:
Thanks for taking the new instructions for such a thorough test drive @legoktm! Back to you.

legoktm (approved):
Awesome! :shipit:

docs/admin/install/install.rst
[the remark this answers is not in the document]
zenmonkeykstop: It is unlikely to be clear to everybody - some folk will take it literally and look for a text menu item, some folk may not be clear as to which settings gear it means (another one literally labeled "Settings" is shown on the left, that applies to single VMs). Specificity is worth it in this case even if it means being a little more verbose.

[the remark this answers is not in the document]
legoktm: Would `**Q > Settings Gear (left sidebar) > Other > Xfce Terminal**` be sufficient?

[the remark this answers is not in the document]
legoktm: Kev also mentioned that when you open the Q menu, there are two settings gears, one in the left sidebar, and one on the right side, as the "Settings" for whichever AppVM is selected by default, so we do need to disambiguate them somehow.

docs/admin/install/prepare.rst
```diff
@@ -75,6 +107,7 @@ After the disk is unlocked and Qubes starts, you will be prompted to complete th
 
 On the configuration screen, ensure that the following options are checked:
 
+ - Default Template should be set to "Fedora 40 Xfce"
```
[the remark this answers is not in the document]
zenmonkeykstop: 4.2.2 is imminent, we can probably chill

docs/admin/install/prepare.rst
zenmonkeykstop: far point.

zenmonkeykstop (approved):

PR #229, opened 2024-07-09
Improve troubleshooting instructions

### LINKED ISSUE #226: Document `cpufreq=xen:hwp=off` workaround on freezes
I was setting up Qubes 4.2.2rc1 and ran into https://github.com/QubesOS/qubes-issues/issues/8825 again, we should probably document the workaround:

1. Edit the xen command-line on boot via grub to set `cpufreq=xen:hwp=off`, then start up
2. Edit `/etc/default/grub` to set the xen command-line persistently as explained in https://github.com/QubesOS/qubes-issues/issues/8825#issuecomment-2100470543 and the follow-up

legoktm:
* Add `cpufreq=xen:hwp=off` full system freezes workaround
  * Document the workaround for
    <https://github.com/QubesOS/qubes-issues/issues/8825> by setting the
    corresponding Xen command-line parameter by editing the grub menu to be
    able to boot into a non-frozen state, and then the grub config itself to
    persist the workaround.
* Move "Unable to reset PCI device" to Troubleshooting page
  * Despite it only affecting specific hardware, this seems like a better
    place to document it when people run into it during installation.
    The only change to the text itself is that the dom0 terminal instructions
    now refer to the Qubes 4.2 way of opening it.

Fixes #226.

### DIFF
--- a/docs/admin/install/troubleshoot.rst
+++ b/docs/admin/install/troubleshoot.rst
@@ -63,6 +63,38 @@ For both device IDs (e.g. ``dom0:00_1f.6`` and ``dom0:00_14.3``), you will need 
 
 .. |screenshot_sys_net_pci_reset| image:: ../../images/screenshot_sys_net_pci_reset.png
 
+Full system freezes
+~~~~~~~~~~~~~~~~~~~
+
+A `known issue <https://github.com/QubesOS/qubes-issues/issues/8825>`_ with some hardware results in Qubes fully freezing.
+If you encounter this issue, you will need to forcibly restart your computer, usually by holding down the power button.
+
+When you boot up, you will see a black-and-white menu with the following options:
+
+.. code-block:: text
+
+  Qubes, with Xen hypervisor
+  Advanced options for Qubes (with Xen hypervisor)
+  UEFI Firmware Settings
+
+While ``Qubes, with Xen hypervisor`` is selected, press :kbd:`e` to edit the option. You should now see a rudimentary
+edit interface.
+
+Find the line that starts with ``multiboot2   /xen-`` and ends with ``${xen_rm_opts}``. Use the arrow keys to move your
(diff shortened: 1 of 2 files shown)


docs/admin/install/troubleshoot.rst
```diff
@@ -25,6 +25,77 @@ This is a transient error that may affect any of the SecureDrop Workstation VMs.
...
+
+While ``Qubes, with Xen hypervisor`` is selected, press :kbd:`e` to edit the option. You should now see a rudimentary
+edit interface.
+
+Find the line that starts with ``multiboot2   /xen-`` and ends with ``${xen_rm_opts}``. Use the arrow keys to move your
+cursor to before ``${xen_rm_opts}`` and type :kbd:`cpufreq=xen:hwp=off` (leave a space between ``off`` and the ``$``.
+
+Press :kbd:`Ctrl-x` to continue with booting. This will fix the current boot, we now need to make the fix permanent.
+
+Once Qubes has started and you have logged in, open a ``dom0`` terminal via **Q > Terminal Emulator** and type
```
nathandyer, line 89: Should this not be **Q > Settings Gear > Other Tools > Xfce Terminal** now?

nathandyer (changes requested):
One small 4.2-specific nit, but otherwise this looks great to me!

docs/admin/install/troubleshoot.rst
legoktm: Yeah, it should. I just left it as is because the PCI section also needs the same fix and we wanted to try to use the gear emoji in https://github.com/freedomofpress/securedrop-workstation-docs/pull/222 but that doesn't work yet.

Will fix.

rocodes (dismissed):
<3

nathandyer (dismissed):

legoktm:
Rebased.

nathandyer (approved):

PR #231, opened 2024-07-11
Remove uninstall instructions for documentation

### LINKED ISSUE #227: Remove --uninstall from documentation
Currently `--uninstall` mostly works, but is largely a developer tool and not supported for end users. So we shouldn't be mentioning it in documentation, like https://workstation.securedrop.org/en/latest/admin/install/troubleshoot.html#uninstalling-securedrop-workstation

legoktm:
Uninstalling isn't supported, just available as a developer tool.

Fixes #227.

### DIFF
--- a/docs/admin/install/troubleshoot.rst
+++ b/docs/admin/install/troubleshoot.rst
@@ -25,22 +25,6 @@ This is a transient error that may affect any of the SecureDrop Workstation VMs.
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 Transient network issues may cause an installation to fail. To work around this, verify that you have a working Internet connection, and re-run the ``sdw-admin --apply`` command.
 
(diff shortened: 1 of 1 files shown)


rocodes (approved):

nathandyer:
I'm a little conflicted about this one, to be honest. I definitely agree that uninstalling is not a supported user-facing feature, and is really only there as a developer tool. That said, it _is_ something that is available on every SDW installation, and I kind of think it's important to document that the submission key and the config.json file are still left behind in dom0 unless they are manually removed.

legoktm:
> That said, it _is_ something that is available on every SDW installation, and I kind of think it's important to document that the submission key and the config.json file are still left behind in dom0 unless they are manually removed.

OK, that's a reasonable point. Let me propose an alternative in a bit.

PR #238, opened 2024-07-15
Update general availability timeline

deeplow:
- general availability will be after 1.0.0
- general availability will be on the second half of 2024 and not mid-2024


legoktm (approved):
Thanks

PR #239, opened 2024-07-15
Check ISO matches .DIGESTS

deeplow:
Adds a step to confirm that the ISO matches the DIGESTS. Assuming the user didn't follow the Qubes OS verification link, under the previous scenario, a compromised ISO could still be malicious, even if the DIGESTS do match.

### DIFF
--- a/docs/admin/install/prepare.rst
+++ b/docs/admin/install/prepare.rst
@@ -33,7 +33,8 @@ Follow the linked instructions to `verify the ISO <https://www.qubes-os.org/secu
 .. code-block:: sh
 
   gpg --keyserver-options no-self-sigs-only,no-import-clean --fetch-keys https://keys.qubes-os.org/keys/qubes-release-4.2-signing-key.asc
-  gpg -v --verify Qubes-R4.2.2-x86_64.iso.DIGESTS && sha256sum -c Qubes-R4.2.2-x86_64.iso.DIGESTS || echo FAILED VERIFICATION
+  gpg -v --verify Qubes-R4.2.2-x86_64.iso.DIGESTS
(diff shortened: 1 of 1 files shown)


docs/admin/install/prepare.rst
```diff
@@ -33,7 +33,7 @@ Follow the linked instructions to `verify the ISO <https://www.qubes-os.org/secu
 .. code-block:: sh
 
   gpg --keyserver-options no-self-sigs-only,no-import-clean --fetch-keys https://keys.qubes-os.org/keys/qubes-release-4.2-signing-key.asc
-  gpg -v --verify Qubes-R4.2.2-x86_64.iso.DIGESTS
+  gpg -v --verify Qubes-R4.2.2-x86_64.iso.DIGESTS && sha256sum -c Qubes-R4.2.2-x86_64.iso.DIGESTS || echo FAILED VERIFICATION
```
legoktm, line 36: I don't think we need the separate echo? It already outputs "FAILED" if it's invalid:

```
$ sha256sum -c Qubes-R4.2.2-x86_64.iso.DIGESTS.bad 
Qubes-R4.2.2-x86_64.iso: FAILED
sha256sum: WARNING: 20 lines are improperly formatted
sha256sum: WARNING: 1 computed checksum did NOT match
```

So my suggestion is:
```suggestion
  gpg -v --verify Qubes-R4.2.2-x86_64.iso.DIGESTS
  sha256sum -c Qubes-R4.2.2-x86_64.iso.DIGESTS
```

legoktm (changes requested):
Nice catch

legoktm (approved):

PR #243, opened 2024-07-15
Set a minimum of 32GB of RAM

### LINKED ISSUE #210: Set 32GB as the minimum RAM amount for SecureDrop Workstation

legoktm:
Advertise production requirements here; dev requirements can go in the dev docs.

Fixes #210.


cfm (approved):
Good call.

Issue #230, opened 2024-07-11
Consider using `:menuselection:` role


legoktm:
https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-menuselection

turns
```
:menuselection:`Start --> Programs`
```
into 

"Start ‣ Programs"

would make the Qubes menu parts just a bit nicer.

PR #250, opened 2024-07-22
Fix a few menu invocations & dark mode issue

### LINKED ISSUE #221: Update docs for Qubes 4.2

### LINKED ISSUE #249: Keyboard formatting is invisible in dark mode
The new troubleshooting instructions in https://workstation.securedrop.org/en/latest/admin/install/troubleshoot.html render as invisible text in dark mode:

docs excerpt showing the problem

eloquence:
- Fixes a few more menu references. I've standardized the text as much as possible to make it easier to search & replace in future.
- Fixes #249
- Resolves #221

## Checklist
- [x] Locally linted & artisanally  previewed

### DIFF
--- a/docs/admin/install/troubleshoot.rst
+++ b/docs/admin/install/troubleshoot.rst
@@ -31,7 +31,7 @@ On some hardware, network devices (Ethernet and Wi-Fi) will not immediately work
 
 |screenshot_sys_net_pci_reset|
 
-Open a ``dom0`` terminal via **Q > Settings Gear > Other Tools > Xfce Terminal**, and run the following command to list the devices connected to the ``sys-net`` VM.
+Open a ``dom0`` terminal via **Q > Gear Icon (left-hand side) > Other Tools > Xfce Terminal**, and run the following command to list the devices connected to the ``sys-net`` VM.
 
 .. code-block:: sh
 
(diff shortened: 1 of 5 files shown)


docs/admin/reference/backup.rst
```diff
@@ -24,7 +24,7 @@ Preserve files from ``dom0`` and ``sd-gpg``
 
 Preserve configuration files and private key material by copying them into ``dom0``.
 
-In a ``dom0`` Terminal via **Q ▸ Terminal Emulator**:
+In a ``dom0`` Terminal via **Q > Gear Icon (left-hand side) > Other Tools > Xfce Terminal**:
```
eloquence, line 27: The ▸ is arguably nicer, but it was only used in this chapter, so I've replaced it for easier search & replace across the docs. If folks prefer, I can do a global search & replace to use ▸ instead (it may be time to start a style guide ..).

docs/admin/install/troubleshoot.rst
```diff
@@ -31,7 +31,7 @@ On some hardware, network devices (Ethernet and Wi-Fi) will not immediately work
 
 |screenshot_sys_net_pci_reset|
 
-Open a ``dom0`` terminal via **Q > Settings Gear > Other Tools > Xfce Terminal**, and run the following command to list the devices connected to the ``sys-net`` VM.
+Open a ``dom0`` terminal via **Q > Gear Icon (left-hand side) > Other Tools > Xfce Terminal**, and run the following command to list the devices connected to the ``sys-net`` VM.
```
eloquence, line 34: As was discussed in #228, unfortunately there are two settings gears visible at the same time in the menu, so we have to disambiguate.

docs/admin/reference/backup.rst
nathandyer: I agree the ▸ is preferable to the > symbol, although I'd be curious if any others on the team have strong opinions one way or the other. For now, I agree that consistency is the most important - thanks for catching.

And a big +1 to a style guide (especially if we can enforce via linting/CI)

nathandyer (approved):
LGTM! I'll go ahead and tag this for the stable version once it merges, especially considering the current dark mode issue.

docs/admin/reference/backup.rst
legoktm: #230 would make it easier to use the unicode arrow symbols.

PR #273, opened 2024-10-25
Update hardware recommendations to include qubes-certified list

### LINKED ISSUE #211: Phase out specific hardware recommendations

zenmonkeykstop:
Draft mode for now - it needs:
 - [ ] up-to-date sections on FPF-tested laptops, basically  the most modern 3 testing laptops in use across the dev team.
 - [x] A blurb about the Qubes HCL with pros and cons of it as a resource
 - [ ] Validation that we're ok with this approach to recommendations.

Fixes #211 
 
Updates hardware reccos:
- recommends qubes-certified, FPF-tested, and HCL options, in that order
- moves config instructions for Thinkpad BIOS updates out of hardware reccos page, to make the point of the page clearer.


zenmonkeykstop:
(I would expect the laptops listed under FPF-tested to change based on what folks currently have, eg newer X1s, frameworks etc)

docs/admin/reference/hardware.rst
```diff
@@ -52,118 +60,25 @@ The 2nd-generation ThinkPad T14 **with an 11th-generation Intel Core processor**
...
-
-  Once complete, remove the USB.
-
-- Plug the USB into the ThinkPad.
-
-- Boot the ThinkPad and follow the prompts to enter its startup and boot menus, likely via the <Enter> and <F12> keys, respectively.
-
-- Follow the on-screen instructions to update the BIOS, including any mandatory reboots. Note that the instructions may refer to an update CD instead of your update USB.
-
-USB-C ports
```
nathandyer, line 162: Was the decision to remove the section on USB-C ports because this information is no longer applicable, or because it's too in-the-weeds?

docs/index.rst
```diff
@@ -56,6 +56,7 @@ against malware and other security risks. It is built on Qubes OS and requires a
    admin/reference/troubleshooting_updates
    admin/reference/provisioning_usb
    admin/reference/backup
+   admin/reference/thinkpad_bios
```
nathandyer, line 59: For the sake of consistency, we should probably either have a small guide for Framework BIOS updates as well, or at least link out to their knowledgebase article on the topic: https://knowledgebase.frame.work/framework-laptop-bios-releases-S1dMQt6F

nathandyer:
Thanks @zenmonkeykstop, even in early drafts this is looking really great. I think this is the correct approach, and I like the way you have explained everything. I left a couple small in-line comments just to jot down some thoughts.

I'm a bit iffy on whether or not it makes sense to include the note about Heads, but it is worthwhile for folks to know it's available, so it's probably reasonable to include it where you did.

Overall, I think this is a huge improvement! Would it be helpful if I tacked on a commit to fill some of the missing sections?

docs/index.rst
nathandyer: Er, that link was for the 11th gen Intel board specifically. Looks like each product has its own BIOS update page. We'd either need to point to https://knowledgebase.frame.work/categories/products_en-r1aUmxtrd, or consider linking to each individual page in the aforementioned small guide.

zenmonkeykstop:
> Thanks @zenmonkeykstop, even in early drafts this is looking really great. I think this is the correct approach, and I like the way you have explained everything. I left a couple small in-line comments just to jot down some thoughts.
> 
> I'm a bit iffy on whether or not it makes sense to include the note about Heads, but it is worthwhile for folks to know it's available, so it's probably reasonable to include it where you did.

Yup that part in particular is worth debating. IMO it's worth mentioning because certified machines often have it as an option, but we could go either way.
 
> Overall, I think this is a huge improvement! Would it be helpful if I tacked on a commit to fill some of the missing sections?

But of course!

docs/index.rst
zenmonkeykstop: My pref would be to link wherever possible - if that's an option I'd link from the section in the hardware reccos. Where there's a long or involved process that we have to document locally, I'd prefer to have it in a separate page.

docs/admin/reference/hardware.rst
zenmonkeykstop: Bit of both - it references 8th-gen T480s so n/a, and if it is caused by bios settings in install section it can probably be a note there.

nathandyer:
Thanks again! I just tacked on a commit which expands the Framework 13 section with a short explanation, and links out to the BIOS upgrade guide for the 13-gen CPUs. It also describes the Qubes HCL, and lists some of the benefits and downsides. I'm sure there are some that I missed, so we can add/remove to those as we see fit.

I also added a tiny commit that we can remove if we want to drop it, which just takes out the AMD processor recommendation since I know there has been some iffyness in how well they work in Qubes.

I left the X1 and T14 sections as-is for now; I think they're roughly still in an OK position, but we can tweak those as needed.

docs/admin/reference/hardware.rst
```diff
@@ -52,118 +60,56 @@ The 2nd-generation ThinkPad T14 **with an 11th-generation Intel Core processor**
...
+* It tells you exactly what is and isn't working within the system, so you can decide if a device you own will function well enough to suit your needs
+* Devices get tested across many different configurations and Qubes versions
 
-- Boot the ThinkPad and follow the prompts to enter its startup and boot menus, likely via the <Enter> and <F12> keys, respectively.
+However, there are some things to consider:
 
-- Follow the on-screen instructions to update the BIOS, including any mandatory reboots. Note that the instructions may refer to an update CD instead of your update USB.
+* Reports are not verified for their accuracy by either the Qubes team or Freedom of the Press Foundation
+* Reports don't automatically get tested with newer Qubes releases
+* Reports may reflect the experience on older Qubes versions, rather than the current version (where hardware support may have improved)
```
eloquence, line 105: Suggest refactoring thse two bullet points:

* Reports correspond to a specific Qubes OS version, and may not reflect breaking changes or expanded  hardware support in the most recent Qubes OS version

docs/admin/reference/hardware.rst
```diff
@@ -52,118 +60,56 @@ The 2nd-generation ThinkPad T14 **with an 11th-generation Intel Core processor**
...
-- Download the file called either **BIOS Update (Bootable CD)** or **BIOS Update (Utility & Bootable CD)**.
+After Qubes starts for the first time, when ``sys-net`` fails to start, follow the troubleshooting instructions for :ref:`reset_pci`, but only for the ``dom0:00_1f.6`` Ethernet device.
 
-.. note::
-  A Tails USB can be used for the verification and conversion process described below, but the Lenovo support site blocks requests over Tor, preventing the ISO download. To work around this, either:
+.. _framework_13_series:
 
-  - download the BIOS ISO on a different computer and transfer it to Tails using a USB stick, or
-  - download the ISO in Tails using the Unsafe Browser as follows:
+Framework 13 (13th-generation)
```
eloquence, line 73: Since it's the only recommendation that doesn't require hardware-specific fixes, maybe we should make it the first one in the list?

eloquence:
Draft looks good from my POV, left a couple comments.

docs/admin/reference/hardware.rst
nathandyer: That makes sense to me - I've pushed an updated commit with this change in place

nathandyer:
@zenmonkeykstop This can probably be flipped out of draft mode now, unless there are any other preliminary items you think we need to address first.

nathandyer:
@deeplow It's definitely possible to add a link like that with Sphinx/RST/RTD. Historically I believe we have intentionally avoided doing so, as to not recommend a particular source for purchase in hopes of diversifying where orgs are getting their hardware. It might be good to continue that precedent here (Frameworks are available directly from the manufacturer, but if there are any resellers or other vendors, it may be good for said diversity). But I have no strong feelings, and am happy to go whichever direction the team prefers.

nathandyer (dismissed):
This PR still looks good to me, and I'm inclined to go ahead and merge it. It has been reviewed by @eloquence and @deeplow in an unofficial capacity, and the only lingering question was whether to link out to vendors (which appears to have shaken out for sake of consistency to _not_ linking out to vendors, to encourage a wider diversity and avoid potential supply-side issues).

We should further expand this with more hardware as we can, and we can always revisit the decision whether to explicitly link out, but I'm going to go ahead and approve/merge for the sake of making this available now to orgs that need it.

nathandyer (approved):

PR #275, opened 2024-10-30
Add introductory note crediting key open source components

eloquence:
If this kind of language works, could add similar language to main SD docs. I probably forgot some important shout-outs though!

### DIFF
--- a/docs/general/introduction.rst
+++ b/docs/general/introduction.rst
@@ -33,4 +33,29 @@ and viewing. SecureDrop Workstation combines all of those steps
 into one workflow on one machine: a Qubes computer that
 combines the *Journalist Workstation* and the *Secure Viewing Station*.
 
-For more information on SecureDrop Workstation, see our :doc:`faq <../journalist/faq>`.
+Who is behind SecureDrop Workstation?
+-------------------------------------
+SecureDrop and SecureDrop Workstation are open source projects of
+`Freedom of the Press Foundation (FPF) <https://freedom.press/>`_, a
+US-based nonprofit organization. You can support our work through
+`your donation <https://freedom.press/donate>`_ or by
+`contributing to SecureDrop development <https://developers.securedrop.org/en/latest/contributing.html>`_.
+
+Our work would not be possible without the larger open source community.
+
+The foundation of SecureDrop Workstation is `Qubes OS <https://www.qubes-os.org/>`_.
+FPF has directly sponsored Qubes OS development, and we encourage you to
+`donate to Qubes OS <https://www.qubes-os.org/donate/>`_ as well.
+
+SecureDrop Workstation uses `Whonix <https://www.whonix.org/>`_ to maintain
+connectivity with the Tor network. You can `make a donation to the Whonix project <https://www.whonix.org/wiki/Donate>`_.
+
+In addition, SecureDrop Workstation relies on other open source projects such as
+`grsecurity <https://www.grsecurity.net>`_, `Python <https://www.python.org/>`_,
+`Debian <https://www.debian.org/>`_, `Fedora <https://fedoraproject.org/>`_,
+`GNOME <https://www.gnome.org/>`_, `GnuPG <https://gnupg.org/>`_,
+`LibreOffice <https://www.libreoffice.org/>`_,
+`Audacious <https://audacious-media-player.org/>`_, and others. Please consider
+contributing to these projects as well.
+
+For more information on SecureDrop Workstation, see our :doc:`FAQ <../journalist/faq>`.
--- a/docs/general/introduction.rst
+++ b/docs/general/introduction.rst
@@ -15,8 +15,6 @@ Broadly speaking, this means that even if files in one of your virtual machines
 are exposed to malware, files in others still have some protection, which is
 not true of other operating systems.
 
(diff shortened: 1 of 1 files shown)


docs/general/introduction.rst
```diff
@@ -33,4 +31,32 @@ and viewing. SecureDrop Workstation combines all of those steps
...
+Our work would not be possible without the larger open source community.
+
+The foundation of SecureDrop Workstation is `Qubes OS`_.
+FPF has directly sponsored Qubes OS development, and we encourage you to
+`donate to Qubes OS <https://www.qubes-os.org/donate/>`_ as well.
+
+SecureDrop Workstation uses `Whonix <https://www.whonix.org/>`_ to maintain
+connectivity with the Tor network. You can `make a donation to the Whonix project <https://www.whonix.org/wiki/Donate>`_.
+
+In addition, SecureDrop Workstation relies on other open source projects such as
```
rocodes, line 51: I slightly worry that, if we get into the level of granularity of (for example) packages installed in SDW VMs, we will leave out some folks inadvertently and prioritize others. (For example, do we list only user-facing applications? Do we list all libraries and components? Do we list PyQt? etc)

What I would propose as a first step is that we highlight and link to QubesOS first, as you have done, then have a section mentioning other projects on whom we have a substantive reliance and/or have engaged with directly (with donation or contribution links as the case may be) with wording like "please consider supporting these projects with your time, with a financial contribution if you are able, or by following their work" or something (not all take donations).

Then we can either stop at that with a line about how the components of SD ecosystem are open source and it's important to support the community, or collect a longlist of other projects we want to mention (but I think that will take some time and be a bit more ambiguous).

I suggest:

QubesOS

then (alphabetical)

Localization Lab https://www.localizationlab.org/donate
Sequoia-PGP https://sequoia-pgp.org/contribute/ (not a donation link)
Weblate https://weblate.org/en/donate
Whonix https://www.whonix.org/wiki/Donate

Then optionally we can also list the major other projects (Debian, Fedora, Python, etc), but I feel a little dicey about that because we get far into "who do we include". I would love a second opinion (maybe @legoktm has thoughts..?)

docs/general/introduction.rst
```diff
@@ -33,4 +31,32 @@ and viewing. SecureDrop Workstation combines all of those steps
...
 combines the *Journalist Workstation* and the *Secure Viewing Station*.
 
-For more information on SecureDrop Workstation, see our :doc:`faq <../journalist/faq>`.
+Who is behind SecureDrop Workstation?
+-------------------------------------
+SecureDrop and SecureDrop Workstation are open source projects of
+`Freedom of the Press Foundation (FPF) <https://freedom.press/>`_, a
+US-based nonprofit organization. You can support our work through
+`your donation <https://freedom.press/donate>`_ or by
+`contributing to SecureDrop development <https://developers.securedrop.org/en/latest/contributing.html>`_.
```
rocodes, line 40: A nit: maybe we can flip the order here and say "you can support our work by contributing to our projects [link], making a financial donation [link]," and (maybe?) adding "or following and sharing the work we do (newsletter/Mastodon link)"?.

I also prefer the "contributing to our projects" wording slightly over the "contributing to development," just to make sure the translators etc feel included :)

rocodes (changes requested):
Hi @eloquence - thanks so much for preparing this! I have left a suggestion about the list of projects we call out explicitly, with a couple options (keep it short, or really make a longlist). I think I'm in favour of a thoughtful shortlist so that we don't get into the "who's in, who's out" territory. Let me know your thoughts. :)

docs/general/introduction.rst
legoktm: I think the "big" projects are a good idea to start with at least (including Debian/Fedora/Python and Tor too). I'd be okay skipping Qt since (AFAICT) it's a for-profit company.

Once we have some SBOMs we can go into more detail and maybe have a comprehensive list across server (flask, etc.) + workstation + client, but that shouldn't hold this up.

eloquence: Thank you both for the review. I've taken another pass at this, let me know what you think! 

I've added Localization Lab, Weblate and Sequoia as part of this change.

In terms of software running in VMs, I've relied on https://github.com/freedomofpress/securedrop-workstation/blob/main/tests/vars/sd-viewer.mimeapps to enumerate individual viewer apps. Most of these are part of the GNOME project; the ones that I individually called out are not.

I am in favor of including these, because they are so essential for the day-to-day user experience of the SecureDrop Workstation, and because the existence of an allow-list does seem to make it practical to maintain an enumerated list.

At the same time, I think it is defensible to not enumerate _every_ transitive library, dependency, or build system component, because such a list would be impractical for us to maintain. Instead, I've included a general comment acknowledging the reality of these deeper dependencies, and encouraging folks to direct financial contributions and effort wherever it can make a positive difference.

docs/general/introduction.rst
eloquence: I've flipped the order here and now use the more general "contributing" without specifying "development". I've not added the "follow/share news" just to keep it focused mainly on the core message of acknowledging the open source ecosystem we rely on.

rocodes (approved):
Thanks @eloquence for the changes and the writeup, very appreciated

PR #278, opened 2024-11-20
Add multi-delete docs

rocodes:
Ready for review

- Remove section about sources with no reply keys (deprecated)
- Add instructions on multi-delete and top/bottom status bar


docs/journalist/sources.rst
```diff
@@ -50,21 +56,17 @@ client.
...
 
-Deleting a conversation
------------------------
+Deleting a single conversation
+''''''''''''''''''''''''''''''
 
-You can delete a source conversation by clicking on the three dots at the top
-righthand side of the application window, beside the timestamp. You should
+You can delete a single source conversation by clicking on the three dots at
+the top righthand side of the application window, beside the timestamp. You should
```
nathandyer, line 66:
```suggestion
the top right-hand side of the application window, beside the timestamp. You should
```

docs/journalist/sources.rst
```diff
@@ -16,6 +16,12 @@ Once logged in, you will see a chat-like user interface consisting of three pane
   with a specific source (messages, files, and journalist replies) will be
   displayed here.
 
+- The top panel holds a toolbar that allows you to apply actions in bulk.
```
nathandyer, line 19: Wondering if "bulk" is a meaningful term here. I feel like if I were to encounter this organically, it might not make sense to me. It's also the only time the term is used here in the docs.

Maybe?:

```suggestion
- The top panel holds a toolbar that allows you to apply actions to multiple items at once.
```

rocodes:
Thanks for review @nathandyer - have added screenshots and addressed your comments. 
Unfortunately (because of what I was mentioning before) you can see I have a dev vm open (because of `make dev-tor`) so they aren't truly prod-quality screenshots, but in the interests of :shipit: I think it's okay?

nathandyer (approved):
Approving after a re-review following changes made from requested changes during draft mode.

Issue #95, opened 2021-12-21
Add "passphraseless submission key" to prerequisites


eloquence:
While our key generation docs suggest that it is safe to generate a passphraseless submission key, they do not strictly require it. It's certainly possible to use an SVS with a passphrase-protected key.
https://docs.securedrop.org/en/stable/generate_submission_key.html

Our SDW installation procedure and test plans do not currently account for a key protected by a passphrase. Until/unless this is a use case we explicitly support, I would suggest that we list a passphraseless submission key as a prerequisite for the SecureDrop Workstation.

conorsch:
> Our SDW installation procedure and test plans do not currently account for a key protected by a passphrase. Until/unless this is a use case we explicitly support [...]

Given that [split-gpg does not support passphrases on privkeys](https://www.qubes-os.org/doc/split-gpg/#current-limitations), we're not likely to change support anytime soon. Agree that clarification in the docs is the right approach.

eaon:
That split-gpg doc is outdated then :smile: I use split-gpg with pinentry-gnome3 every day! But even so, the automated use from salt during initial setup that would make the use of passphrases on private keys cumbersome to support

conorsch:
> That split-gpg doc is outdated then

Good to know! I was surprised to see the feature explicitly omitted, even though I don't use it myself. And the original point stands: we're not likely to support it on SDW in the foreseeable future.

zenmonkeykstop:
This came up recently for a new user, it would be good to update the install docs to flag the need for a passphrase0less version of the submission privkey. I'm agnostic on whether or not to provide instructions on *removing* the key, but I probably would not add said instructions in the main install flow as it would be irrelevant to a large majority of users.

nathandyer:
Just to add a couple thoughts here:

- This might be a good candidate for a `.. note::`
- I recommend that we suggest that users reach out to support in the event that they have a submission key with a passphrase set so we can assist with exporting a version without a passphrase set. I'm not sure that the removal instructions are widely useful enough to live in our main docs (but happy to be persuaded otherwise)

zenmonkeykstop:
If possible, the docs should allow folks to function without falling back to support. I'd agree they're probably not going to be widely useful (coz current server setup recommends against a passphrase) but putting them in a linked page makes sense IMO.

PR #284, opened 2024-12-10
[DRAFT] Passhpraseless submission key requirement

lilyannehall:
add a note and reference doc on the lack of support for passphrase protected submission keys and how to remove the passphrase - #95

### DIFF
--- a/docs/admin/install/install.rst
+++ b/docs/admin/install/install.rst
@@ -4,7 +4,12 @@ Installing SecureDrop Workstation
 Copy the submission key
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-In order to decrypt submissions, your SecureDrop Workstation will need a copy of the secret key from your SecureDrop instance's SVS. To protect this key and preserve the air gap, you will need to connect the SVS USB to a Qubes VM with no network access, and copy it from there to ``dom0``. Note that you cannot directly copy and paste to the ``dom0`` VM from another VM - instead, follow the steps below to copy the file into ``dom0``:
+In order to decrypt submissions, your SecureDrop Workstation will need a copy of the secret key from your SecureDrop instance's SVS.
+
+.. note::
+   The copy of your SecureDrop submission secret key `cannot be protected with a passphrase <https://www.qubes-os.org/doc/split-gpg/#current-limitations>`_. To export a copy that does not require a passphrase, see :doc:`/admin/reference/removing_gpg_passphrase`.
+
+To protect this key and preserve the air gap, you will need to connect the SVS USB to a Qubes VM with no network access, and copy it from there to ``dom0``. Note that you cannot directly copy and paste to the ``dom0`` VM from another VM - instead, follow the steps below to copy the file into ``dom0``:
 
 - First, use the network manager widget in the upper right panel to disable your network connection. These instructions refer to the ``vault`` VM, which has no network access by default, but if the SVS USB is attached to another VM by mistake, this will offer some protection against exfiltration.
 
--- a/docs/admin/reference/removing_gpg_passphrase.rst
+++ b/docs/admin/reference/removing_gpg_passphrase.rst
@@ -1,0 +1,27 @@
+Removing the Passphrase from a GPG Key
+======================================
+
+First, note your secret key's ID. You can find it by running the command:
+
+.. code-block:: sh
+
+  gpg --list-secret-keys --keyid-format=long
+
+This will show a list of keys known to GPG. Find the key you wish to remove the passphrase from - the ID will be in place of ``XXXX`` in the output example below.
+
+.. code-block:: sh
+
+  /home/username/.gnupg/secring.gpg
+  ----------------------------------
+  sec   4096R/XXXX <creation date>
+  uid                  name <email.address>
+  ssb   4096R/YYYY <creation date>
+
+Using the key ID in place of ``XXXX``, open the edit key dialogue by running:
+
+.. code-block:: sh
+
+    gpg --edit-key XXXX
+
+GPG will display information about the key and a prompt. Type ``passwd`` into the prompt and press enter. You will be asked for the current passphrase, type it and press enter. When asked for the new passphrase, simply leave the prompt blank and press enter again. Depending on the version of GPG and your desktop environment, you may receive a warning. You can dismiss this and proceed anyway.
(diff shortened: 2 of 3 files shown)


docs/admin/install/install.rst
```diff
@@ -4,7 +4,12 @@ Installing SecureDrop Workstation
 Copy the submission key
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-In order to decrypt submissions, your SecureDrop Workstation will need a copy of the secret key from your SecureDrop instance's SVS. To protect this key and preserve the air gap, you will need to connect the SVS USB to a Qubes VM with no network access, and copy it from there to ``dom0``. Note that you cannot directly copy and paste to the ``dom0`` VM from another VM - instead, follow the steps below to copy the file into ``dom0``:
+In order to decrypt submissions, your SecureDrop Workstation will need a copy of the secret key from your SecureDrop instance's SVS.
+
+.. note::
+   The copy of your SecureDrop submission secret key `cannot be protected with a passphrase <https://www.qubes-os.org/doc/split-gpg/#current-limitations>`_. To export a copy that does not require a passphrase, see :doc:`/admin/reference/removing_gpg_passphrase`.
```
zenmonkeykstop, line 10: I think that doc is out-of-date, the limitation for us is not that it can't work with split-gpg, but with our provisioning script (also users would get prompted for said passphrase quite a bit). Could drop that, or replace it with a link to the securedrop server docs about provisioning the Submission Key that speak to passphrases.

docs/admin/reference/removing_gpg_passphrase.rst
```diff
@@ -0,0 +1,27 @@
+Removing the Passphrase from a GPG Key
```
zenmonkeykstop, line 1: These instructions need better context - they should be performed on dom0 and in a temporary GPG homedir that gets nuked after they're complete. I had some notes in a recent related support issue if that helps.

docs/admin/reference/removing_gpg_passphrase.rst
```diff
@@ -0,0 +1,27 @@
...
+======================================
+
+First, note your secret key's ID. You can find it by running the command:
+
+.. code-block:: sh
+
+  gpg --list-secret-keys --keyid-format=long
```
zenmonkeykstop, line 8: Would need updating with a --homedir, and in general (tho not for this command) probably a --pinentry=loopback for any command with a passphrase prompt.

docs/admin/reference/removing_gpg_passphrase.rst
```diff
@@ -0,0 +1,27 @@
...
+  uid                  name <email.address>
+  ssb   4096R/YYYY <creation date>
+
+Using the key ID in place of ``XXXX``, open the edit key dialogue by running:
+
+.. code-block:: sh
+
+    gpg --edit-key XXXX
+
+GPG will display information about the key and a prompt. Type ``passwd`` into the prompt and press enter. You will be asked for the current passphrase, type it and press enter. When asked for the new passphrase, simply leave the prompt blank and press enter again. Depending on the version of GPG and your desktop environment, you may receive a warning. You can dismiss this and proceed anyway.
```
zenmonkeykstop, line 26: The new password will be prompted for twice, even if blank - the instructions should be Qubes- and dom0-specific, so you can speak to the particular errors/warnings you'd see there.

zenmonkeykstop (changes requested):

docs/admin/install/install.rst
lilyannehall: can i check my understanding real quick - the scenario in which a user would need to remove the passphrase from the submission key is when they decide to enter a passphrase here at steps 9 and 10? https://docs.securedrop.org/en/stable/admin/installation/generate_submission_key.html#create-the-key

and if so, should the language change in that doc too? from "it's okay to skip passphrase" to "do not enter a passphrase"?

zenmonkeykstop: IMO the language there is OK - the next step also assumes it. But we could tweak it. I guess the tension there is that for the SVS it could work either way (and we have no visibility by default into how folks set them up), but for SDW it's now more of a requirement. Either way these docs would need to reflect the no-passphrase requirement.

rocodes:
I took a look, and I think all of the requested changes have already been addressed. I'm going to resolve conflicts and squash + sign commits, but don't think much more is needed

rocodes:
I've resolved conflicts and made a couple small edits, and also tested on Qubes.
- the callout to the extra instructions is placed inline in the install instructions where the user will encounter them: after the initial import of the key from Tails, and if `sdw-admin --validate` fails.  If a user is successful at importing a passphrase-protected key, they'll have just seen the pinentry prompt (from the vault vm, which does have a pinentry program), and that's their reminder to unset the password on the key file.
- `sdw-admin validate` in the extra instructions won't work if the user is in the middle of configuring their setup (no tor hidserv stuff yet), so I've clarified that.

docs/admin/install/install.rst
```diff
@@ -59,7 +64,7 @@ SecureDrop Workstation connects to your SecureDrop instance's API via the *Journ
 
 - Connect the USB drive to a USB port on the Qubes computer, then use the devices widget in the upper right panel to attach it to the ``vault`` VM. There will be 3 listings for the USB in the widget: one for the base USB, one for the Tails partition on the USB, labeled ``Tails``, and a 3rd unlabeled listing, for the persistent volume. Choose the third listing.
 
-- In the the ``vault`` file manager, select the persistent volume's listing in the lower left sidebar. It will be named ```N GB encrypted``, where N is the size of the persistent volume. Enter the persistent volume passphrase to unlock and mount it. When prompted, select the option to **Forget password immediately**.
+- In the the ``vault`` file manager, select the persistent volume's listing in the lower left sidebar. It will be named ``N GB encrypted``, where N is the size of the persistent volume. Enter the persistent volume passphrase to unlock and mount it. When prompted, select the option to **Forget password immediately**.
```
rocodes, line 68: (driveby fixup, extra backtick in ```N GB encrypted``)

eloquence:
@rocodes Looks like the :gear: icon is causing the PDF build to fail, might be quickest to just replace it with a text placeholder for now.

zenmonkeykstop (approved):
LGTM

PR #287, opened 2024-12-19
Include note about LVFS and HP firmware woes in hardware recs

### LINKED ISSUE #285: HW recs: ensure LVFS support
From team meeting, filing a brief ticket: 

In hw recomendations, make sure that in addition to Qubes HCL support, a further restriction is that recommended laptops are well-supported in LVFS so that qubes-fwupd can work.  (Note: This excludes HP laptops).

rocodes:
Fixes #285


nathandyer (approved):
I think this is good for a quick note/first pass.

We probably want to add this/expand it in the hardware guide itself; I also wish we knew if there were other vendors than HP who have similar challenges with firmware.

LGTM, thank you!

PR #292, opened 2025-01-10
Ignore some links that can't be checked in CI

legoktm:
* Skip all .onion URLs since Tor isn't set up
* The GNOME website returns a 403 from cloud networks, including the one GitHub Actions runs from.
* Same with the Lenovo support website, except it just times out.


cfm (approved):
Thanks for adding this, @legoktm.  It will be great to have these false positives go away.

PR #300, opened 2025-02-06
docs: clarify key-update procedure

cfm:
- [x] Clarify commands and expected output in code blocks
- [x] Add explicit verification step

## Testing

- [ ] None of the "Run command" `code-blocks` end with ` `` ` left over from their inline format.
- [ ] The verification step in 57712a399f42cae9c27b8cd8cf4310ce888893c3 is reasonable and works.


zenmonkeykstop (approved):
LGTM, thanks @cfm

PR #308, opened 2025-03-24
Recommend using exFAT for VeraCrypt drives

legoktm:
Now that we support it (c.f. https://github.com/freedomofpress/securedrop-workstation/issues/1267), recommend it to users since it supports larger files compared to FAT.


nathandyer (approved):
LGTM - not sure that exFAT provides much benefit in a world where we limit files to 500MB, but then again, it's widely supported so no reason _not_ to

Issue #312, opened 2025-04-22
Update printer docs (Driverless Printing Release)


deeplow:
There are a couple of printer-related changes that need to go in with the client 0.16.0 release:
- Add new printer troubleshooting workflow - client no longer reports most printing issues. They should handled through the print dialog (which conveniently indicates problems with the printer or through the printer tray icon, which shows up when a file is being printed). Additional troubleshooting advanced avenues such as [ipp-usb quirks](https://github.com/OpenPrinting/ipp-usb/blob/master/ipp-usb-quirks/README) could be mentioned as resolution possibilities (mostly linking to possibilities and resources -- but getting another printer is probably the easiest)
- #283  (references throughout the docs)

nathandyer:
Thanks @deeplow - will work on this early next week with @ChumOfChance, and tag you for a review once it's ready

PR #315, opened 2025-04-29
Printer recommendation update

ChumOfChance:
Updates printer recommendation and references throughout documentation in line with driverless printing support, as per [https://github.com/freedomofpress/securedrop-workstation-docs/issues/312]( https://github.com/freedomofpress/securedrop-workstation-docs/issues/312)


docs/journalist/submissions.rst
```diff
@@ -44,12 +44,11 @@ computer.
...
-To print a document, click the **Print** button. Currently, printing is only
-supported with a specific HP printer model, and for security reasons
-you are required not to use a printer that has any wireless capabilities.
+To print a document, a :doc:`compatible printer <../admin/reference/hardware.html>`  must be plugged into the computer's USB port.
 
-You should have access to a supported printer that has been set up by your
-administrator. The printer must be plugged into the computer's USB port.
+1. Click "Print" button and wait for ``sd-devices`` VM to start.
+2. You will prompted to attach your printer.
+3. A typical Print Document dialog will appear, from which you can configure different print options before printing the document.
```
nathandyer, line 51:
```suggestion
3. A Print Document dialog will appear, from which you can configure different print options before printing the document.
```
I think we can probably drop the word 'typical' here, just to try and avoid intimidating new users who might find it a bit unfamiliar

docs/admin/reference/hardware.rst
```diff
@@ -125,5 +125,15 @@ Qubes compatible, the HCL is a good choice.
...
-In order to print submissions, a supported non-networked printer is required. We have tested and recommend the HP LaserJet Pro M404n. More printer options will be added in future releases.
+SecureDrop Workstation only supports printing over USB, and so it is essential to choose a compatible printer that **has a USB port**. 
 
+Driverless IPP printing is used to support a large selection of modern printers. Compatible printers can be easily identified by their support for the Apple AirPrint or Moipra standards:
+
+.. figure:: images/airprint.jpg
+
+.. figure:: images/moipra.jpg
+
+You may consult Apple's `list of printers that support AirPrint <https://support.apple.com/en-us/HT201311#printers>`_, Moipra's `list of certified products <https://mopria.org/certified-products>`_, or OpenPrinting's `list of printers supporting driverless printing <https://openprinting.github.io/printers/>`_, looking for a compatible printer that will print over USB.
```
nathandyer, line 136:
```suggestion
To find a printer that supports driverless IPP printer over USB, you may consult Apple's `list of printers that support AirPrint <https://support.apple.com/en-us/HT201311#printers>`_, Moipra's `list of certified products <https://mopria.org/certified-products>`_, or OpenPrinting's `list of printers supporting driverless printing <https://openprinting.github.io/printers/>`_.
```
Recommended rearranging this sentence for clarity

nathandyer (changes requested):
This is looking great! Thanks @ChumOfChance! I left two in-line suggestions. This is also my first time hearing about Moipra - so I have a bit more learning to do about this!

Once those suggestions are in (if they're agreeable), I'll tag a few other reviewers just to get some extra feedback.

docs/admin/reference/hardware.rst
nathandyer: Ah, that's an interesting distinction! In that case I think you're right that my suggestion is not an improvement.

There is something about the original wording that my brain has a hard time wrapping around. Perhaps if we changed it to something like "..., being careful to choose a model that provides a USB port" rather than "looking for a compatible printer that will print over USB."

This is obviously a loose suggestion, and I defer to you as well as the other reviewers.

legoktm:
IMO whenever y'all are satisfied, feel free to merge it, we won't tag a new stable docs release until post-release.

legoktm:
> Oh. Great. I was working under the assumption that the "latest" docs were literally `main`. Good to know.

That is correct though, there are two published sets of docs:
* https://workstation.securedrop.org/en/stable/ -> corresponds to `stable` tag and what we send to newsrooms to use, manually updated during release time or whenever some other docs update merits it
* https://workstation.securedrop.org/en/latest/ -> corresponds to `main` branch

Because we have the separation, it's fine stage things on main even if it hasn't been released yet :)

legoktm:
(sorry, hit the wrong button)

nathandyer (approved):
Just re-reviewed the PR since it has undergone significant changes from my first review. Really loving the shape it has ended up taking. Thanks @ChumOfChance for such thorough work, and thanks for everyone's thoughtful feedback here in the thread while iterating through it.

I think this is in a mergeable state - will do so, with the caveat that we will not forward the stable tag until these changes are shipping.

PR #316, opened 2025-05-01
Addition and reorganization of BIOS update instructions

ChumOfChance:
This PR incorporated unused changes to BIOS update instructions from @nathandyer 's https://github.com/freedomofpress/securedrop-workstation-docs/pull/224/commits/67ef692473cd7ace8903ce010fd6e243aeb71d8e

The previous ThinkPad-specific BIOS instructions page is removed in favor of a new page including those specific instructions and the more generalized instructions.

BIOS instructions on the Recommended Hardware page are removed in favor of reference to the above new page. 

I left the external links to the model-specific BIOS instructions to the two Framework models we recommend in their place on the Recommended Hardware page.


docs/admin/reference/bios_update.rst
```diff
@@ -1,8 +1,43 @@
...
+BIOS Update Instructions
+====================================
+
+.. _general_BIOS_update:
+
+
+Automatic BIOS Updates
+----------------------
+
+These instructions should work for many recent laptops, include the two ThinkPad models specifically included in our :doc:`hardware`.
```
nathandyer, line 10:
```suggestion
These instructions should work for many recent laptops, including the two ThinkPad models specifically included in our :doc:`hardware`.
```

Tiny typo

nathandyer (changes requested):
This is looking great! One small fix requested, and then I suggest we request another reviewer just for another set of eyes

PR #336, opened 2025-08-07
Reflect updater showing up on first reboot

deeplow:
The preflight-updater always shows even on the first reboot  after installing the workstation. This changes the language to reflect this fact and moves the desktop icon to a note.

## Checklist

This change accounts for:
- [ ] local preview of changes beyond typo-level edits

### DIFF
--- a/docs/admin/install/install.rst
+++ b/docs/admin/install/install.rst
@@ -216,7 +216,7 @@ This command will take a considerable amount of time and approximately 4GB of ba
 Test the Workstation
 ~~~~~~~~~~~~~~~~~~~~
 
-The preflight updater will start automatically after logging into the system. Since the system should be up-to-date, no updates this step should be fast. Please follow the preflight updater's instructions. 
+The preflight updater will start automatically after logging into the system. Please follow the preflight updater's instructions. 
 
   .. note::
 


docs/admin/install/install.rst
```diff
@@ -216,7 +216,11 @@ This command will take a considerable amount of time and approximately 4GB of ba
 Test the Workstation
 ~~~~~~~~~~~~~~~~~~~~
 
-To start the SecureDrop Client, double-click the SecureDrop desktop icon that was set up by the previous command. The preflight updater will start and check for updates. The system should be up-to-date and no updates should be required, but if updates are available follow the instructions in the preflight updater to apply them.
+The preflight updater will start automatically after logging into the system. Since the system should be up-to-date, no updates this step should be fast. Please follow the preflight updater's instructions.
```
eloquence, line 219: Stray words: "no updates this step"

But I would maybe kill that sentence altogether. The user has no reference point for "fast" at this point. Given all the Salt shenanigans, VM restarts, network bounce, etc., they may still feel it's pretty slow.

docs/admin/install/install.rst
```diff
@@ -216,7 +216,11 @@ This command will take a considerable amount of time and approximately 4GB of ba
 Test the Workstation
 ~~~~~~~~~~~~~~~~~~~~
 
-To start the SecureDrop Client, double-click the SecureDrop desktop icon that was set up by the previous command. The preflight updater will start and check for updates. The system should be up-to-date and no updates should be required, but if updates are available follow the instructions in the preflight updater to apply them.
+The preflight updater will start automatically after logging into the system. Since the system should be up-to-date, no updates this step should be fast. Please follow the preflight updater's instructions. 
+
+  .. note::
+
+    There is also a SecureDrop desktop icon to start the SecureDrop Client, in case it is ever closed.
```
eloquence, line 223: Instead of the passive voice, how about:

> If you close the SecureDrop Client during your session, you can launch it again using the SecureDrop icon on the desktop.

eloquence (approved):

PR #343, opened 2025-09-10
Qubes Icons

ChumOfChance:
In preparing documentation changes for [SecureDrop 1.4.0](https://github.com/freedomofpress/securedrop-workstation/issues/1419) to account for the shift away from `sd-whonix`, I noticed a few references to the [Qubes App menu](https://github.com/QubesOS/qubes-desktop-linux-menu) and the [Qubes Domains menu](https://www.qubes-os.org/doc/getting-started/#user-interface) which were overly confusing. This is in part no doubt due to changes in Qubes over the years (many of which are not even properly referenced in the Qubes docs).

I added icons for these two menus, and updated all the menu references to use them. I also swapped `>` for the `▸` character in menu paths because it's nicer. 

## Checklist

This change accounts for:
- [x] local preview of changes beyond typo-level edits


eloquence:
With ALT text, I think the icon approach is much clearer than trying to target similar visual elements via text descriptions. I agree an icon would be nice for the gear references as well. We tried using :gear: before but IIRC that caused problems with the LaTeX rendering for PDFs. If inline images work, great!

PR #369, opened 2026-01-29
Document installing and using the new SecureDrop App

### LINKED ISSUE #365: Document installing the new SecureDrop Application
In conjunction with work on the [developer documentation related to the new SecureDrop Application](https://github.com/freedomofpress/securedrop-client/issues/2862), we should go ahead and document the installation and basic use instructions for the app. This will be useful to share with folks ahead of the upcoming audit (and of course, will be applicable for the wider application rollout later this year)

nathandyer:
Fixes #365

This is a larger set of changes that documents installing and using the new SecureDrop app, and builds a separate version of the docs from the 'app' branch (which is accessible in the stable/latest/app version switcher).

Note: I seem to be unable to actually test the switcher functionality locally (even the "known good" config seems broken for me). So I'd encourage folks to pay extra attention there when reviewing.

## Test plan

- [ ] Visual review
- [ ] CI passes
- [ ] Using the new install instructions for switching to the app results in a working app for systems that are set to "staging" or "dev"
- [ ] The version switcher has a new option called "app"
- [ ] Clicking the "app" option in the switcher actually changes to the new app

## Checklist

This change accounts for:
- [x] local preview of changes beyond typo-level edits

### DIFF
--- a/docs/admin/install/install.rst
+++ b/docs/admin/install/install.rst
@@ -158,7 +158,7 @@ Once the update check is complete, the SecureDrop Client will launch. Log in usi
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 By default, you will receive the SecureDrop Client, our original tool for journalists to access the sources,
-messages, and attachments within SecureDrop. Our newest tool, the SecureDrop App, can be enabled manually during the initial roll-out period. After this period is complete, the SecureDrop App will become the new default.
+messages, and attachments within SecureDrop. Our newest tool, [the SecureDrop App](https://github.com/freedomofpress/securedrop-client/tree/main/app#readme), can be enabled manually during the initial roll-out period. After this period is complete, the SecureDrop App will become the new default.
 
 If you would like to switch to the SecureDrop App immediately, you can follow these steps:
 


docs/journalist/sources.rst
ChumOfChance: I don't think this section is accurate for the App. Messages from different journalists appear to be in the same colour, and are identified by their username.

docs/general/known_issues.rst
```diff
@@ -20,22 +20,13 @@ for instructions on how to privately report it.
 Current known issues
```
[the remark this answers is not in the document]
legoktm: I don't have any off the top of my head that are different from the client's known issues.

docs/admin/reference/troubleshooting_connection.rst
```diff
@@ -182,13 +182,13 @@ If this does not resolve the issue, please contact us for assistance.
 Customizing Synchronization Timeouts
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The SecureDrop Workstation Client application performs a synchronization on
+The SecureDrop App performs a synchronization on
 launch, which may time out in situations where there are a large number of
```
legoktm: We can delete the caveat about timing out.

legoktm (commented):
Nice! See inline comments

docs/journalist/sources.rst
nathandyer: (left in for now)

PR #385, opened 2026-05-13
Fix the smallest of typos

apyrgio:
Nothing says "did my part" more than adding an extra word, but still :shrug:

### DIFF
--- a/docs/general/introduction.rst
+++ b/docs/general/introduction.rst
@@ -18,12 +18,12 @@ not true of other operating systems.
 What is SecureDrop Workstation?
 -------------------------------
 
-SecureDrop Workstation combines a native SeucreDrop client application — SecureDrop Inbox — with Qubes OS, to produce a system that faster and simpler for journalists to use.
+SecureDrop Workstation combines a native SecureDrop client application — SecureDrop Inbox — with Qubes OS, to produce a system that is faster and simpler for journalists to use.
 
 A key feature of SecureDrop is that journalists can receive submissions from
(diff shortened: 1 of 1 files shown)


docs/general/introduction.rst
```diff
@@ -18,12 +18,12 @@ not true of other operating systems.
 What is SecureDrop Workstation?
 -------------------------------
 
-SecureDrop Workstation combines a native SeucreDrop client application — SecureDrop Inbox — with Qubes OS, to produce a system that faster and simpler for journalists to use.
+SecureDrop Workstation combines a native SeucreDrop client application — SecureDrop Inbox — with Qubes OS, to produce a system that is faster and simpler for journalists to use.
```
deeplow: Thanks!! The hilarious bit is that there was slightly bigger one right under our noses. If you could, please also add this one 🙌 — it's really hard to spot, but if one looks close enough 🔎.

deeplow (approved):

docs/general/introduction.rst
apyrgio: The utility of this PR is increasing manifold!

deeplow:
<strike>@nathandyer we may need your stamp of approval here, when you're back.</strike> (**update**: never mind. Turns out it was missing signatures).

apyrgio:
I just saw that signatures are required, so I force-pushed with mine.

deeplow:
OK. Then that was the missing bit.

PR #398, opened 2026-06-10
Remove all existing references to the client

### LINKED ISSUE #394: Remove legacy client references
Following release 1.7.1, the "classic" SecureDrop Client is now officially deprecated. Its references need to be removed from the docs.

I found [this one](https://github.com/freedomofpress/securedrop-workstation-docs/blob/cf84d23748d4922d5ed93f60d61c5f1881c77743/docs/journalist/starting_client.rst#falling-back-to-the-legacy-securedrop-client), but there may be more.

nathandyer:
This PR removes all references to SecureDrop Client, as it has officially reached end of life.

Fixes #394

## Test plan
- [ ] Visual review

## Checklist

This change accounts for:
- [ ] local preview of changes beyond typo-level edits

### DIFF
--- a/docs/admin/install/install.rst
+++ b/docs/admin/install/install.rst
@@ -150,9 +150,9 @@ The preflight updater will start automatically after logging into the system. Pl
 
   .. note::
 
-    If you close the SecureDrop Client during your session, you can launch it again using the SecureDrop icon on the desktop. 
+    If you close the SecureDrop Inbox during your session, you can launch it again using the SecureDrop icon on the desktop. 
 
-Once the update check is complete, the SecureDrop Client will launch. Log in using an existing journalist account and verify that sources are listed and submissions can be downloaded, decrypted, and viewed.
+Once the update check is complete, SecureDrop Inbox will launch. Log in using an existing journalist account and verify that sources are listed and submissions can be downloaded, decrypted, and viewed.
 
 .. _Password Management Section:
 
--- a/docs/general/known_issues.rst
+++ b/docs/general/known_issues.rst
@@ -35,7 +35,7 @@ Current known issues
   the Tails-based *Secure Viewing Station*. :doc:`Broader file type support is planned <supported_filetypes>`.
 - If the *Submission Key* for your SecureDrop server was rotated in the past,
   you must manually re-add the old key to your `sd-gpg` VM in order to
-  view old submissions in SecureDrop Client. `Contact Support <https://securedrop.org/help/>`_ for assistance.
+  view old submissions in SecureDrop Inbox. `Contact Support <https://securedrop.org/help/>`_ for assistance.
 - We do not support uninstalling SecureDrop Workstation without wiping all of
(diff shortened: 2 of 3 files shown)


jskinne3 (approved):
Searching the codebase, this looks to me like a complete removal of references to the old SD Client 👍

deeplow:
The only one I could spot is in the file name itself: `docs/journalist/starting_client.rst`. Do we want to keep this one so we don't break redirects?

ChumOfChance:
@deeplow I've been thinking about that file name too. We can change it and change internal links/references in the docs. It only might cause some link rot if there are any external places that link directly to that page... which is probably unlikely?

deeplow:
I think we should always assume some link somewhere will get broken (even something like internal newsroom admin notes). Supposedly readthedocs [supports redirects](https://docs.readthedocs.com/platform/stable/user-defined-redirects.html#redirecting-a-page).

ChumOfChance:
Internal notes is a fair point. Combining the workstation and regular docs as part of https://github.com/freedomofpress/securedrop-docs/issues/749 is already going to change a lot of file names/locations where certain information is found in the docs (see the `main` branch of https://github.com/freedomofpress/securedrop-docs). So I think either now, or during that restructure, changing the name of `starting_client.rst` should happen.

legoktm:
> I think we should always assume some link somewhere will get broken (even something like internal newsroom admin notes). Supposedly readthedocs [supports redirects](https://docs.readthedocs.com/platform/stable/user-defined-redirects.html#redirecting-a-page).

If sphinx can't do it, then we can ask infra to set up redirects for us (they already have a pile of them from past renames/restructuring).

nathandyer:
> Internal notes is a fair point. Combining the workstation and regular docs as part of [freedomofpress/securedrop-docs#749](https://github.com/freedomofpress/securedrop-docs/issues/749) is already going to change a lot of file names/locations where certain information is found in the docs (see the `main` branch of https://github.com/freedomofpress/securedrop-docs). So I think either now, or during that restructure, changing the name of `starting_client.rst` should happen.

I've been thinking about this thread for a bit, and I believe waiting for the restructure makes the most sense.

Once we complete the docs restructure, there won't _be_ a Sphinx documentation site at workstation.securedrop.org, and _everything_ will need to redirect to the main docs site. So implementing the redirects at that level (with infras help) rather than using the options within Sphinx is probably the more future-proof way to go.

And since this will be happening in the relatively short term (once admin tools on Qubes lands), I think it probably makes sense to _not_ rename that file yet, but handle all the wonkiness from the rename as part of that larger transition.

My preference would be to go ahead and merge this as-is, then, if the reviewers are OK with that.

deeplow:
Merging, per @jskinne3's approval and conclusion to defer the redirects. I've added a comment [to the docs restructure issue](https://github.com/freedomofpress/securedrop-docs/issues/749#issuecomment-4767335427) in order for that to be followed up on.

